Issue #2, opened 2016-02-24
supporting more properties via cfstore 


shroffk:
Currently cfstore supports
hostName
iocName
time
pvStatus

upgrade 1.
The python channlefinder library already consists of a python script which populates properties  by parsing pv names using regular expressions defined in a configuration file.

upgrade 2.
the cfstore plugin should parse the info associated the record itself and attach the appropriate tags/properties and properties. Thus ioc engineer can control the population of information into channelfinder from the ioc configuration files.

shroffk:
refere to issue #14

Issue #1, opened 2016-02-24
cfstore plugin requirements


shroffk:
We have been working on developing a plugin for recsync which pipes the data to channelfinder.

Currently the features supported are 
1. which ioc startup, all the records are added to channelfinder along with properties iocName, hostName, time, and ovStatus
2. upon ioc reboot, the "time" 

Missing
When an IOC shuts down the information available in the transaction object consists of only the host ip (as part of the TR.src). Since there might be more than one ioc running on a particular host, the IOCNAME information is necessary to update channelfinder correctly.

mdavidsaver:
> When an IOC shuts down the information available in the transaction object consists of only the host ip

Both IP address and port number are included, which together are unique.

shroffk:
Currently cfstore dose not record the port, instead it uses the iocname+hostname.
Maybe the port should be stored as a property too.

mdavidsaver:
Keeping "ip:port" seems reasonable

PR #3, opened 2016-05-10
Cf store

shroffk:
Address the following issues
https://github.com/ChannelFinder/recsync/issues/1
https://github.com/ChannelFinder/recsync/issues/2


server/recceiver/cfstore.py
```diff
@@ -0,0 +1,207 @@
...
+    
+        '''
+        Currently using the hostIP and the iocPort
+        if 'IOCNAME' in TR.infos:
+                iocName = TR.infos['IOCNAME']
+        if 'HOSTNAME' in TR.infos:
+                hostName = TR.infos['HOSTNAME']
+        '''
+        if 'ENGINEER' in TR.infos:
+            owner = TR.infos['ENGINEER']
```
mdavidsaver: What are the consequences if ENGINEER is set to something like "mdavidsaver x3698"?

shroffk:
I have tested the following processes on a local development setup
1. ran recsync-server with cf.conf
2. started an IOC and verified that the 4 cf-store properties are added with correct time stamp
3. shut down the IOC, ensured that all channels were marked "InActive" with correct timestamp
4. added new properties and tags to channels, restarted the IOC to ensure that cfstore does not touch the properties not owned by it.

server/recceiver/cfstore.py
shroffk: the owner name in channelFinder will be set to "mdavidsaver x3698", the down the pipe consequences would depend on the cf AA setup.

server/recceiver/cfstore.py
```diff
@@ -0,0 +1,207 @@
...
+    '''
+    ch = {u'name':chName,u'owner':chOwner,u'properties':[]}
+    if hostName != None:
+        ch[u'properties'].append({u'name' : 'hostName', u'owner' : chOwner, u'value': hostName})
+    if iocName != None:
+        ch[u'properties'].append({u'name' : 'iocName', u'owner' : chOwner, u'value': iocName})
+    if pvStatus:
+        ch[u'properties'].append({u'name' : 'pvStatus', u'owner' : chOwner, u'value': pvStatus})
+    if time:
+        ch[u'properties'].append({u'name' : 'time', u'owner' : chOwner, u'value': time})
```
mdavidsaver, line 172: The duplicate code in `updateChannel()` and `createChannel()` should be consolidated.  It could also be simplified using the builtin function `locals()`

``` python
VS = locals()
for V in ('hostName', 'iocName', 'pvStatus', 'time'):
    if V in VS:
        ch[u'properties'].append({u'name' : V, u'owner' : chOwner, u'value': VS[V]})
```

server/recceiver/cfstore.py
```diff
@@ -0,0 +1,207 @@
...
+        self.running = 1
+        print "CF_START"
+        try:
+            '''
+            Using the default python cf-client.
+            The usr, username, and password are provided by the channelfinder._conf module.
+            '''
+            self.client = ChannelFinderClient()
+        except:
+            print 'Failed to create cf client'
```
mdavidsaver: Swallowing unrecoverable exceptions like this is not good.  Should be re-raised.

server/recceiver/cfstore.py
mdavidsaver: Presumable this will fail for nsls2 or frib as "mdavidsaver x3698" is not a valid user name?  This variable has traditionally been free-form, I'd suggest adding a new CF specific variable (eg. "CF_USERNAME").  At least don't fail hard if "ENGINEER" isn't a valid username.  Maybe retry with the default.

mdavidsaver:
@mskinner5278 fyi

server/recceiver/cfstore.py
shroffk, line 172: +1, the current code seems like left over from the old cf-update where there was significant difference in how to update/create channel

server/recceiver/cfstore.py
shroffk: This was mistakenly committed, I have reverted the code

mdavidsaver:
I've made some additional (imo) minor changes.  @mskinner5278 and/or @shroffk please comment, then I will merge.

shroffk:
+1 from me

PR #5, opened 2016-05-26
System configuration clobbers local configuration

dxmaxwell:
The configuration specified on the command line with -f option get overridden by the global configuration.


mdavidsaver:
Looks like I have the order of `paths = ['~/.recceiver.conf', '/etc/recceiver.conf']` backwards as well.

PR #6, opened 2016-06-01
Installation support

dchabot:
Rename installed conf file.
Permit local overrides of logfile location.


mdavidsaver:
No problem with `$LOG_FILE`.  I should have done this initially.

Renaming the default/example config file is fine as it isn't so useful.  However, some care should be taken as debian has a troublesome default of failing package installation if a daemon doesn't start with the default config.  Avoid with `dh_installinit --error-handler=true`.

mdavidsaver:
I should note that with debian 8 w/ systemd this no longer seems to be an issue, but with sysv it is (I've been bitten).

PR #7, opened 2016-06-02
Fix optional dependency on Channel Finder client library.

dxmaxwell:
I initially used the same approach as you to make the cfstore plugin optional, but I ran into a problem with the Twisted Plugin cache.

If you install recceiver WITHOUT the CF client library installed the Twisted plugin cache will be refreshed at the time of installation and the cfstore plugin will not be available. Then if you install the CF client library and try to use the cfstore plugin it will not be available without manually triggering a refresh of the Twisted plugin cache.  

As a result, I implemented an alternative approach which allows the cfstore plugin to be properly cached at the time of installation, but does not require the CF client to be installed until the plugin is actually started.

### DIFF
--- a/server/recceiver/cfstore.py
+++ b/server/recceiver/cfstore.py
@@ -6,7 +6,6 @@ from zope.interface import implements
 from twisted.internet import defer
 from twisted.application import service
 from twisted.enterprise import adbapi as db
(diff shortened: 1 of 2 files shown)


mdavidsaver:
Right, this is better.

PR #15, opened 2016-10-03
Cfstore 3.0.0

shroffk:
Preparation branch for testing a PR to the main recsync project

### DIFF
--- a/server/recceiver/cfstore.py
+++ b/server/recceiver/cfstore.py
@@ -5,6 +5,8 @@ _log = logging.getLogger(__name__)
 
 from zope.interface import implements
 from twisted.application import service
+from twisted.internet.threads import deferToThread
+import time
 import interfaces
 import datetime
 
@@ -15,7 +17,7 @@ import datetime
 # delrec = a set() of records which are being removed
 # infos = dictionary of client infos
 # recinfos = additional infos being added to existing records 
-# "recid: {key:vale}"
+# "recid: {key:value}"
 #
 
 
@@ -26,9 +28,12 @@ class CFProcessor(service.Service):
 
     def __init__(self, name, conf):
         _log.info("CF_INIT %s", name)
-        self.name,self.conf = name,conf
-        
-        
+        self.name, self.conf = name, conf
+        self.channel_dict = dict()
+        self.iocs = dict()
+        self.client = None
+        self.currentTime = getCurrentTime
+
     def startService(self):
         service.Service.startService(self)
         self.running = 1
@@ -36,8 +41,9 @@ class CFProcessor(service.Service):
         from channelfinder import ChannelFinderClient
         # Using the default python cf-client.
         # The usr, username, and password are provided by the channelfinder._conf module.
-        self.client = ChannelFinderClient()
-        
+        if self.client is None:  # For setting up mock test client
+            self.client = ChannelFinderClient()
+
     def stopService(self):
         service.Service.stopService(self)
         #Set channels to inactive and close connection to client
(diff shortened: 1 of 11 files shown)


shroffk:
@dylan171 @mdavidsaver 
Reviewed most of the changes...still not sure about a few things due to my lack of expertise...

My last commit added a few major changes.
1. I replaced certain individual CF calls with a single one, this should result in hundreds of fewer n/w calls depending on the number of channels.
2. IOC developers can add info tags with the syntax 
   info (properties, "propertyName=propertyValue,...")
   e.g.
   `info(properties, "device=$(M)")`

server/recceiver/cfstore.py
```diff
@@ -69,15 +70,49 @@ def commit(self, transaction_record):
...
+            if "properties" in  recinfos:
+                if rid in pvInfo:
+                    recProperties = recinfos["properties"]
+                    properties = []
+                    for prop in recProperties.split(","):
+                        p = prop.strip().split("=")
+                        properties.append({u'name': p[0], u'owner': owner, u'value': p[1]})
+                    pvInfo[rid]["infoProperties"] = properties
+                else:
+                    _log.exception("could not find the associated record for properties")
```
mdavidsaver: `.exception()` is only useful when called when there is an active exception (eg. within a `catch:` block).  Otherwise just use `.error()`.

server/recceiver/cfstore.py
```diff
@@ -69,15 +70,49 @@ def commit(self, transaction_record):
...
+        _log.debug(pvInfo)        
+            
+        pvNames = [info["pvName"] for rid, (info) in pvInfo.iteritems()]
+        
+        delrec = list(TR.delrec)
+        _log.info(delrec)
```
mdavidsaver: Should have some prefix string

server/recceiver/cfstore.py
```diff
@@ -86,17 +121,21 @@ def __commit__(self, TR):
         for pv in delrec:
             if iocid in self.channel_dict[pv]:
                 self.channel_dict[pv].remove(iocid)
-                self.iocs[iocid]["channelcount"] -= 1
+                if iocid in self.iocs:
+                    self.iocs[iocid]["channelcount"] -= 1
```
mdavidsaver, line 125: Is the added test here from paranoia, or due to an issue encountered?  I'm a bit nervous about the manual ref. counter, and would like to be clear if there is any issue discovered.

server/recceiver/cfstore.py
```diff
@@ -134,133 +173,101 @@ def dict_to_file(dict, iocs, conf):
...
+        existingChannels[ch["name"]] = ch
+    
+    for pv in new:        
+        newProps = [{u'name': 'hostName', u'owner': owner, u'value': hostName},
+                     {u'name': 'iocName', u'owner': owner, u'value': iocName},
+                     {u'name': 'iocid', u'owner': owner, u'value': iocid},
+                     {u'name': 'pvStatus', u'owner': owner, u'value': "Active"},
+                     {u'name': 'time', u'owner': owner, u'value': iocTime}]
+        infoProperties = [info["infoProperties"] for rid, (info) in pvInfo.iteritems() if info["pvName"] == pv and "infoProperties" in info ]
+        _log.debug(infoProperties)
```
mdavidsaver: same here.

mdavidsaver (approved):
Overall looks good.  A few minor points about logging, please fix if you have time, otherwise I will eventually.

server/recceiver/cfstore.py
shroffk: I am not completely comfortable with this either. I did have an issue when doing my tests (restarting the same ioc on the same machine multiple times) which resulted in this check...
The main goal of my changes is to minimize the network calls and to allow the parsing of info tags in the iocs

shroffk:
@dylan171  @mdavidsaver  can we merge this by the end of this week?

dxmaxwell:
Just deployed the latest cfstore for testing and ran into this error:

2016-11-03 15:22:16-0400 [CastReceiver,0,10.41.200.35] DEBUG:urllib3.connectionpool "GET /channelfinder//resources/channels?~name=_FE_LEBT%3ARFC1_D0987%3Areconnected%7C_FE_LEBT%3ARFC2_D0987%3Areconnected%7C_FE_LEBT%3ARFC2_D0987%3AfwdAmplDev%7CFE_L
EBT%3ARFC1_D0987%3AHLIM_RD_CAV%7Ciocllrf%3AAUTOSAVE%3ASR_3_State%7CFE_LEBT%3ARFC2_D0987%3AILK_RSTS_AFHI%7C_FE_LEBT%3ARFC1_D0987%3Aset_pulserOn%7CFE_LEBT%3ARFC1_D0987%3APOS_RSET_TRIG.......
2016-11-03 15:22:16-0400 [CastReceiver,0,10.41.200.35] DEBUG:recceiver.cfstore error: 414 Client Error: Request-URI Too Long

shroffk:
I am just workin on this and had hoped to sneak a fix in before anyone noticed 
:(
With the new cfstore, while I have reduced the number of n/w calls by combining each individula channle calls into a single one....this can significantly reduce the number of times you have to query CF... however if the ioc has many pv's the above error was likely. I apologize for the shoddiness ...

dxmaxwell:
@mdavidsaver, I discuss the ownership of the RecSync repository with @shroffk and we agreed that the ChannelFinder organization would be the best fit right now.

Please merge this PR and then transfer the repository to ChannelFinder. Thanks.

mdavidsaver:
> Please merge this PR and then transfer the repository to ChannelFinder. Thanks.

Step 1 complete.  Some problems with step 2.

```
ChannelFinder/recsync already exists and You don’t have admin rights to ChannelFinder 
```

PR #18, opened 2016-11-09
Initialize reserved items before sending data

mark0n:
This prevents Valgrind from complaining about use of uninitialized data. This fixes FRIB bug [EPICS-97].


mdavidsaver:
I guess this means that your compiler is using larger reads to access the 8/16 bit fields.  Which it is allowed to do.

Issue #19, opened 2017-01-06
Persist arbitrary info tags to CF


dxmaxwell:
The cfstore plugin only supports the 'properties' info tag as described in issue #15.

CfStore should be configurable with a whitelist of other info tags to persist to CF.

@berryma4 @dchabot @shroffk

dchabot:
@shroffk 

Perhaps you can close this via #23?

PR #23, opened 2017-08-18
ENH: infotag whitelist

dchabot:
Bump @mdavidsaver @shroffk.

Addresses #19.

While adding arbitrary Properties is kinda sweet, it may be nice to have more administrative control.

Yes. I said that...

See also PR #15.
@dylan171 @berryma4

### DIFF
--- a/server/recceiver/cfstore.py
+++ b/server/recceiver/cfstore.py
@@ -88,6 +88,9 @@ class CFProcessor(service.Service):
         for rid, (rname, rtype) in TR.addrec.iteritems():
             pvInfo[rid] = {"pvName":rname}            
         for rid, (recinfos) in TR.recinfos.iteritems():
+            if recinfos in self.whitelist:
+                # add tag, values to container for shipping to CF
+                pass
             if "properties" in  recinfos:
                 if rid in pvInfo:
                     recProperties = recinfos["properties"]
--- a/server/recceiver/cfstore.py
+++ b/server/recceiver/cfstore.py
@@ -34,6 +34,8 @@ class CFProcessor(service.Service):
     def __init__(self, name, conf):
         _log.info("CF_INIT %s", name)
         self.name, self.conf = name, conf
+        self.whitelist = conf.get('infotags', []).split(' ')
+        print 'whitelisted info tags: {}'.format(self.whitelist)
         self.channel_dict = defaultdict(list)
         self.iocs = dict()
         self.client = None
--- a/server/recceiver/cfstore.py
+++ b/server/recceiver/cfstore.py
@@ -204,7 +204,6 @@ def __updateCF__(client, pvInfo, delrec, channels_dict, iocs, hostName, iocName,
         raise Exception('missing hostName or iocName')
     
     channels = []
-    checkPropertiesExist(client, owner)
     """A list of channels in channelfinder with the associated hostName and iocName"""
     old = client.findByArgs([('iocid', iocid)])
     if old is not None:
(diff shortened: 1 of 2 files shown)


server/recceiver/cfstore.py
```diff
@@ -86,16 +89,22 @@ def __commit__(self, TR):
...
+            recinfo_wl = [p for p in self.whitelist if p in recinfos.keys()]
+            if recinfo_wl:
+                pvInfo[rid]['infoProperties'] = list()
+                for infotag in recinfo_wl:
+                    _log.debug('INFOTAG = {}'.format(infotag))
+                    property = {u'name': infotag, u'owner': owner,
+                                u'value': recinfos[infotag]}
+                    # XXX this could raise
+                    prop_exists = self.client.find(name=pvInfo[rid]['pvName'],
+                                                   property=[(infotag, '*')])
```
mdavidsaver: In addition to possibly throwing, having one round trip per info() will adversely impact performance.

server/recceiver/cfstore.py
```diff
@@ -86,16 +89,22 @@ def __commit__(self, TR):
...
+                    # if it doesn't exist, create placeholder for new property
+                    if not prop_exists:
+                        self.client.set(property={u'name': infotag,
+                                        u'owner': owner})
```
mdavidsaver: Seems to me that this service should own the whitelisted properties as it will manage them.  Then it could create them on startup, avoiding the extra round trips.

server/recceiver/cfstore.py
```diff
@@ -34,6 +34,9 @@ class CFProcessor(service.Service):
     def __init__(self, name, conf):
         _log.info("CF_INIT %s", name)
         self.name, self.conf = name, conf
+        wl = conf.get('infotags', list())
+        self.whitelist = [s.strip(', ') for s in wl.split()] if wl else wl
+        print 'whitelisted info tags: {}'.format(self.whitelist)
```
mdavidsaver: _log.info() instead of print?

mdavidsaver (changes requested):
I had originally thought to "whitelist" by prefix, probably "cf:", which would then be stripped off.

> info("cf:ABC", "XYZ")

Adds property "ABC" with value "XYZ".

From the prospective of code a static whitelist is easier to implement as it permits pre-creation (or at startup) of all allowed names.

My suggestion would be to use a prefix "cf:" to identify which info()s will become properties.  Then have cfstore fetch a list of all property names at startup to know when a new property name needs to be created.

Also, to support a full lifecycle, I think it would be useful to have a way to remove properties from an info() tags.  While this doesn't have to be implemented right away, it might be useful to eg. decide that an empty value, or extra special prefix, will trigger removal.  Reserving this now avoids any possible breakage later on.

server/recceiver/cfstore.py
dchabot: I thought I nuked that `print`, but I guess I missed it.

Unfortunately, the logger is not up and running at this point.
Or, if it is, it's swallowing those particular messages... ;-)

server/recceiver/cfstore.py
dchabot: Agreed.
This was just me feeling my way around in a first pass.

server/recceiver/cfstore.py
dchabot: Good idea.
I'll amend the PR tomorrow.

shroffk:
@dchabot @mdavidsaver 
Can we merge and close this?

I was looking over the 3 reviews
1st seems like it is resolved
2nd describes good design practice but can be made into it's own issue ticket and PR
3rd.. I am not sure about this one

server/recceiver/cfstore.py
```diff
@@ -15,6 +15,7 @@
 import datetime
 import os
 import json
+from channelfinder import ChannelFinderClient
```
mdavidsaver: This import was previously moved into the constructor so that the cfstore.py can be imported when channelfinder isn't present. 

6d990bac2bbd1d199ccbccd0e79c7abcf7a9e1b3

mdavidsaver (commented):
Other than the import issue, looks good.

shroffk:
@dchabot can you cherry pick dylan's commit?
otherwise I can merge this PR and add the cherry picked changes via another PR

server/recceiver/cfstore.py
dchabot: That's dirty...

shroffk:
Ok, I will merge this now, @mdavidsaver we can update the recceiver_plugin.py later

PR #26, opened 2018-02-13
add Python3 support

shengb:
@mdavidsaver @shroffk 
I modified the code, which now supports both Python2 & Python3.
Did not try the minimum version of twisted library it needs.


mdavidsaver:
I don't see any real issues.  About the only thing I would ask is to remove the py2 specific code instead of commenting it out.

shengb:
A quick search shows that dict.items() method might have performance issue in Python2.
Wondering whether should use dict.iteritems() instead.
Keep it as is for now.

PR #31, opened 2018-10-18
Add missing dependency on lsb-base

mark0n:
The init script includes `/lib/lsb/init-functions` which is provided by `lsb-base`. This fixes a Lintian error.


mark0n:
That was quick - thanks Kunal!

PR #32, opened 2018-11-02
Add lsb-base to depends on recceiver binary package

dxmaxwell:
@mark0n 
I still see the 'init.d-script-needs-depends-on-lsb-base' lintian error on the latest build. I think the lsb-base package needs to be added as a runtime depends.


mark0n:
Oops, yes, of course.

PR #33, opened 2018-11-26
Record alias in CF

tanviash:
What is the change for?
->The changes are for adding alias records in CF. Currently, alias are not logged to the CF.

Will this affect everyone?
->No. This feature is optional to have. It can be set to on/off through demo.conf file using keyword 'alias'.

What does this new code change do?
-> For every record with an alias, the code creates a new 'alias record' in Channelfinder. 
For instance, if the below is the record from an IOC that has alias attached
record(longin, "$(P)li") {
  alias("$(P)lix1")
  alias("$(P)lix2")
  info("test", "testing")
}
then, two additional channels 'lix1' and 'lix2' will be created in CF with exactly same properties as their parent channel 'li'. The two alias channels 'lix1' and 'lix2' will have a property called 'alias' whose value is the parent channel name.

This is how they appear in channelFinder.
[{"name":"test:lix1","owner":"myself","properties":[{"name":"alias","owner":"cfstore","value":"test:li"},{"name":"hostName","owner":"cfstore","value":"develop-vmphy0"},{"name":"iocName","owner":"cfstore","value":"myioc"},{"name":"iocid","owner":"cfstore","value":"127.0.0.1:53430"},{"name":"pvStatus","owner":"cfstore","value":"Active"},{"name":"time","owner":"cfstore","value":"2018-11-26 16:32:14.902487"}],"tags":[]},

### DIFF
--- a/server/demo.conf
+++ b/server/demo.conf
@@ -59,7 +59,7 @@ idkey = 42
 
 [cf]
 # cf-store application
-
+#alias = on
 # Debug output file location.
 # Produces no file when not defined.
 # debug_file_loc = /home/devuser/recsyncdata.txt
--- a/server/demo.conf
+++ b/server/demo.conf
@@ -59,7 +59,7 @@ idkey = 42
 
 [cf]
 # cf-store application
-#alias = on
+alias = on
 # Debug output file location.
 # Produces no file when not defined.
 # debug_file_loc = /home/devuser/recsyncdata.txt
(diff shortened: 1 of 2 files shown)


tanviash:
@dylan171 @berryma4 @shroffk @dchabot @mdavidsaver  Please review the pull request. Also let me know if anyone else should be reviewing it.

Issue #27, opened 2018-06-21
Scalability issues


shroffk:
FRIB (@berryma4) reported that in situations where many IOC's each with 10k's of pvs are started at the same time the receiver crashes.
It is due to memory overflows with the receiver bufferring the record lists from all the IOC's.

mdavidsaver:
The main tool I made for bounding peak memory usage is ```maxActive```, which restricts the number of IOCs allowed to concurrently dump record lists.

https://github.com/ChannelFinder/recsync/blob/7ebed83a9870d1343d0fcb359e28d6ffea1848f5/server/demo.conf#L42-L44

@berryma4 Have you tried lower this from the default?  Also, did you save any stack traces from the crashes you saw?

mdavidsaver:
@berryma4 ping.  (maybe @dylan171 knows the story)

anderslindho:
Turns out maxActive never actually was used, or that it had been broken at some point in history - see https://github.com/ChannelFinder/recsync/pull/146/changes/956ced807cc98e2ac0e13b2ea510d327ce0c84a0. I wonder if we now can/should close this ticket?

PR #38, opened 2019-02-11
Fix problem with client RIDs being reused when more then 65536 records

dxmaxwell:
The use of `htons` caused the RID to increment is steps of 65536 and as a resulted limited the number of unique RIDs to 65536.  The problem can be masked if the server create small Transaction objects that contain less than 65536 records.

This should explain at least some of the salability (#27) problems we have been seeing.

@mdavidsaver I tried to find all places where rid is assigned, but I would appreciate your review.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,5 +10,6 @@ dropin.cache
 *.log
 *.pid
 *.db
+client/configure/*.local
 
 *~
--- a/client/.ci/travis-build.sh
+++ b/client/.ci/travis-build.sh
@@ -1,0 +1,14 @@
+#!/bin/sh
+set -x -e
+
+export EPICS_HOST_ARCH=`sh $HOME/epics-base/startup/EpicsHostArch`
+
+make -j2
+
+[ "$TEST" = "YES" ] || exit 0
+
+lspci
+
+make tapfiles
+
+make -s test-results || find . -name '*.tap' -print0 | xargs -0 -n1 prove -e cat -f
--- a/client/.ci/travis-prepare.sh
+++ b/client/.ci/travis-prepare.sh
@@ -1,0 +1,66 @@
+#!/bin/sh
+set -e -x
+
+cat << EOF > configure/RELEASE
+EPICS_BASE=$HOME/epics-base
+EOF
+
+git clone --depth 10 --branch $BASE https://github.com/epics-base/epics-base.git $HOME/epics-base
+
+export EPICS_HOST_ARCH=`sh $HOME/epics-base/startup/EpicsHostArch`
+
+case "$STATIC" in
+static)
+    cat << EOF >> "$HOME/epics-base/configure/CONFIG_SITE"
+SHARED_LIBRARIES=NO
+STATIC_BUILD=YES
+EOF
+    ;;
+*) ;;
+esac
+
+case "$CMPLR" in
+clang)
+  echo "Host compiler is clang"
+  cat << EOF >> $HOME/epics-base/configure/os/CONFIG_SITE.Common.$EPICS_HOST_ARCH
+GNU         = NO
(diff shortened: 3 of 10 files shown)


mdavidsaver:
Copy+paste strikes again...  I'll take a look.

mdavidsaver:
A couple of minor points.  My build with stack-protector turned up a bug triggered by test code only (yes there is test code).  https://github.com/mdavidsaver/recsync/commit/5ba50bb51645b34d20a37c2306f151e65fb64619

This test code needs a corresponding change for the RIDs, which I had consistently wrong.  3a0bb2ee88fa934bda4ecd660e6b295505aa9afe

Also, I made at attempt at travis-ci.org config, by copying another project, but I seem to have confused travis-ci by reusing the origin repo name for a fork, (oops I forgot).  So I haven't checked this yet.  Hopefully I haven't confused github as well  0ca8dc7789c1019c1e0c161ff513270bd8c0b23e

dxmaxwell:
Bump @mdavidsaver What is needed to merge this PR?  I have some server changes ready for a PR, but would prefer to have this merged first.

mdavidsaver:
Will you pull in the changes I mentioned previously to this branch/PR?

dxmaxwell:
Done. The changes to **client/configure/RELEASE** gave me problems in testing and I needed to revert it, is that correct? Looks like a development configuration.

mdavidsaver:
The newer convention is to ship `configure/RELEASE` with only comments.  Users then write a unversioned file `configure/RELEASE.local` with configuration.

Of course you can revert if you like.

dxmaxwell:
I support adopting better conventions, but I'm trying to understand my options. My goal is to have the package built from this repository be compatible with the other epics-base packages. If this was a standard package then I could patch in the RELEASE.local file, but since this is a native package that doesn't work. Another solution is to add the RELEASE.local file to the FRIB internal repository, but that doesn't help anyone else.  

Looks like I can either revert the file, or convert this to a standard Debian package. Any other ideas?

mdavidsaver:
For the package build, the answer is easy.  Write `configure/RELEASE.local` before `make`, then clean it.

https://github.com/epicsdeb/devlib2/blob/master/debian/rules#L9

dxmaxwell:
That works! I've made the change. If it looks complete, please merge it.

dxmaxwell:
@mdavidsaver  I found that I needed to remove RELEASE.local just before the dh_install command because the '--fail-missing' option is used.  I noticed from the build log that EPICS Debhelp already cleans up the RELEASE file during dh_epics_postinstall:
```
22:41:28 make[1]: Leaving directory '/build/recsync-1.3.4+frib6+0~20190321023951.18+debian09~1.gbp67a07f'
22:41:28    dh_epics_postinstall -O--parallel
22:41:28 	rm debian/tmp//usr/lib/epics/configure/RELEASE
22:41:28    dh_epics_installlibs -O--parallel
22:41:28 	install -m 755 -d debian/tmp/usr/lib/x86_64-linux-gnu
22:41:28 	mv debian/tmp//usr/lib/epics/lib/linux-x86_64/libreccaster.so.1 debian/tmp/usr/lib/x86_64-linux-gnu/libreccaster.so.1
22:41:28 	rm -f debian/tmp//usr/lib/epics/lib/linux-x86_64/libreccaster.so
22:41:28 	ln -s ../../../x86_64-linux-gnu/libreccaster.so.1 debian/tmp//usr/lib/epics/lib/linux-x86_64/libreccaster.so
22:41:28    dh_epics_install -O--parallel
22:41:28 	epics-recsync-dev EPICS devel package
```
Perhaps EPICS Debhelper should also remove RELEASE.local if it exists?

PR #42, opened 2019-05-20
Refactor and Doc Append for Recceiver

DanielALS:
Trying recsync at the ALS and hitting issues with a clean elasticsearch DB.
I refactored  the recceiver session start method to fail more gracefully if certain tags/properties were not
present.
Added to the server/README the minimum required properties.

### DIFF
--- a/server/recceiver/cfstore.py
+++ b/server/recceiver/cfstore.py
@@ -55,7 +55,9 @@ class CFProcessor(service.Service):
             Using the default python cf-client.  The url, username, and
             password are provided by the channelfinder._conf module.
             """
-            # TODO: Explain why are we importing CF client here.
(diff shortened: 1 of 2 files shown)


server/recceiver/cfstore.py
```diff
@@ -52,35 +55,47 @@ def startService(self):
             Using the default python cf-client.  The url, username, and
             password are provided by the channelfinder._conf module.
             """
+            # TODO: Explain why are we importing CF client here.
```
mdavidsaver: This module is always imported by `twistd`, even when 'proc = cf' isn't given.  This way the ChannelFinderClient import is only attempted if it is actually needed.

shroffk:
Outdated at this point

PR #43, opened 2019-07-01
Improvements to RecCeiver performance

dxmaxwell:
(From changelog)
  * Send PING message immediately when DONE message is received
  * Commit transactions sequentially and support cancellation
  * Collect and log more performance metrics
  * Remove the extra newline character from the log messages
  * Add support for cfstore commit cancellation
  * Allow the ChannelFinder client size limit to be configured
  * Improve cfstore performance by hashing PV info by PV name instead of RID

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -3,8 +3,12 @@ recsync (1.3.6) unstable; urgency=low
   * Send PING message immediately when DONE message is received
   * Commit transactions sequentially and support cancellation
   * Collect and log more performance metrics
+  * Remove the extra newline character from the log messages
+  * Add support for cfstore commit cancellation
+  * Allow the ChannelFinder client size limit to be configured
+  * Improve cfstore performance by hashing PV info by PV name instead of RID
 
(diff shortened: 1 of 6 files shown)


jeonghanlee:
@shroffk @shengb @mdavidsaver How are you? 

 I've discussed with @dylan171 about this pull request last week, which FRIB has used since then. Does anyone have a chance to look at this pull request?  I've heard they has no issue on this request in order to run RecCeiver service at FRIB.

shroffk:
@dylan171 this looks good.
I just want to confirm if this is backward compatible?

While I do not see anything that looks like it would not work with the older reccaster I just wanted to confirm.

dxmaxwell:
@shroffk Yes, this is backwards compatible with reccaster (no changes to the recsync protocol).  

There are some additional configuration options (checkout the changes to server/demo.conf), but the default behaviour should be unchanged.

jeonghanlee:
@shroffk Thank you for quick reply. We are moving to the latest one soon.

PR #47, opened 2020-11-09
Add record type to channel finder from recceiver

tynanford:
At our site, we have found it will be useful for channel finder to include the PV record type. This information is already sent by reccaster, so it wasn't too difficult to add it to recceiver's push to channel finder.

So far everything is working as expected (which makes me a little suspicious). If this seems like a useful addition for the community I can work with you to test and ensure this is fully working.

### DIFF
--- a/server/cf.conf
+++ b/server/cf.conf
@@ -37,3 +37,6 @@ procs = cf
 [cf]
 # a space-separated list of infotags to set as CF Properties
 #infotags = archive foo bar blah
+# Turn on optional alias and recordType properties
+#alias = on
+#recordType = on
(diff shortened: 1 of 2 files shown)


jeonghanlee:
@shroffk could you review this pull request? After this, I can merge them into the master branch.

shroffk:
I wonder if we should support the addition of this property to be configurable (like the alias)

shroffk:
While I do see the benefit of this property as being rather ubiquitous...I think I am fine with merging this for the time being if it is too much work to make this configurable.

tynanford:
I will look into making it configurable. I don't think it should be that difficult

tynanford:
It's working in my initial tests. Can toggle the recordType property in cf.conf and see recceiver's behavior change. Will do some more tests later today

tynanford:
@shroffk @jeonghanlee This is ready for review now

shroffk (approved):

shroffk:
`reqd_props.add(...`

Thank god for adding this, the creation of a complete new list was driving me crazy

jeonghanlee:
@shroffk thank you for your review. Without you r review, we don't catch this option.

PR #48, opened 2021-02-16
Fix bug in recceiver cfstore function name  __merge_property_lists

tynanford:
This bug was created by me here: 21a85ab74905a9a1b292652800a012fa448d510b

__merge_property_list should've been __merge_property_lists


jeonghanlee:
@shroffk could you review this bug fixed which we introduced?

shroffk (approved):

PR #53, opened 2022-03-20
[WIP] Resolves #52

### LINKED ISSUE #52: Discussion - EPICS description field in reccaster/recceiver
Often time the applications that use channel finder for us also would benefit from accessing the EPICS description field. That information could get passed automatically in reccaster

gabrielpreviato:
This PR targets adding the DESC field of PV's on the CF and in the storage DB.

As discussed in #52 adding the DESC field can be very useful for a lot of scenarios.
There is already an implementation on https://github.com/lnls-sol/recsync/pull/1 , however there are some concerns over that implementation that we need to fix/discuss:
- Should we always send the DESC field information? Descriptions can be big, and thus we should think if this can create a network traffic problem.
- We should add in the server side a flag as default to off for capturing the DESC, so we don't break any running system.

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,8 +216,10 @@ The header will have the form
   <td>ATYPE</td>
   <td>RTLEN</td>
   <td colspan="2">RNLEN</td>
+  <td>RDLEN</td>
   <td>RTYPE</td>
   <td>RNAME</td>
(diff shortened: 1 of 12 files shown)


kasemir:
handled by #54

PR #55, opened 2022-05-11
Add environment variable preference setting for cfstore

tynanford:
Discussion here: https://github.com/ControlSystemStudio/epics-codeathon-2022/discussions/12

This is for step 1

I've split this from step 2 since it is independent from changing reccaster to send the full set of env variables. That can be a separate pull request


kasemir:
Can we merge this? For the failed check, all I see is "This check failed" without detail, can't tell if it matters or is more of a build system issue.

PR #57, opened 2022-05-16
Add setReccasterEnvironmentVars iocsh function for custom env vars

tynanford:
Per discussion here: https://github.com/ControlSystemStudio/epics-codeathon-2022/discussions/12#discussioncomment-2733048

This would add the ability to for a user to send their own environment variables in addition to the envs list here: https://github.com/ChannelFinder/recsync/blob/master/client/castApp/src/dbcb.c#L13

So you could call this in st.cmd:

```
setReccasterEnvironmentVars("SECTOR,BUILDING,CONTACT")
```

A code review and suggestions would be greatly appreciated. Thanks


kasemir:
Combined with the recent addition of `RSRV_SERVER_PORT` https://github.com/epics-base/epics-base/pull/269, this allows us to track the TCP connection info of the IOC in the channel finder, which in turn enables building a 'name server'.

shroffk:
@mdavidsaver could you have a look?

client/castApp/src/dbcb.c
```diff
@@ -66,6 +118,11 @@ static int pushEnv(caster_t *caster)
             casterMsg(caster, "Error sending env %s", envs[i]);
     }
 
+    if(!ret && customEnvsDefinition) {
+        ret = pushCustomEnv(caster);
+        free(customEnvsDefinition);
```
simon-ess: Why is this needed to be freed? Doesn't this provide an issue if the IOC loses contact with the recceiver? Wouldn't a separate recceiver not pick up the custom variables when it is discovered?

tynanford: Yes I think you are right, this variable should be kept exactly for the reason you said

simon-ess:
In general I wonder if a better idea would be to append the variables to the pre-generated list of environment variables to pass along. Thus we could use a syntax more akin to
```
addReccasterEnvironmentVar("SECTOR")
addReccasterEnvironmentVar("BUILDING")
addReccasterEnvironmentVar("CONTACT")
```
This should allow the send to not require special code to handle the extra variables. That said, this may have been discussed at the prior code-a-thon.

(This would also avoid re-parsing the `customEnvsDefinition` each time, but that's probably premature optimisation)

simon-ess:
As per my suggestion above, here is one possible implementation: https://github.com/tynanford/recsync/pull/1

tynanford:
thanks I like the change. Per @mdavidsaver we need to move the iocsh function definition to castinit.c and then we don't have to register `addReccasterEnvironmentVar` separately. I have that working with `setReccasterEnvironmentVars`, can update the branch with that for `addReccasterEnvironmentVar`

tynanford:
@mdavidsaver @simon-ess made progress on placing addReccasterEnvironmentVar in castinit.c (https://github.com/tynanford/recsync/tree/update)

Now I see this memset which clears the caster_t struct during casterInit:

https://github.com/ChannelFinder/recsync/blob/master/client/castApp/src/caster.c#L111

So calling `addReccasterEnvironmentVar` works if called after iocInit but if called before iocInit then thecaster->extra_envs is null. Any tips on how I should go about this?

mdavidsaver:
> Any tips on how I should go about this?

I think that the call to `casterInit()` could be moved out of `casthook()` and into `reccasterRegistrar()` without ill effect.

tynanford:
> I think that the call to casterInit() could be moved out of casthook() and into reccasterRegistrar() without ill effect.

Thanks, that works well.

@mdavidsaver @simon-ess  this PR is ready for a review again

Also BTW, I updated the default env list to include RSRV_SERVER_PORT and PVAS_SERVER_PORT as per our discussion last week. With base R7.0.7 I see RSRV_SERVER_PORT make it's way into CF

client/castApp/src/castinit.c
```diff
@@ -65,13 +67,63 @@ static void casthook(initHookState state)
...
+            errlogSevPrintf(errlogMajor, "Error in memory allocation of extra_envs from %s", __func__);
+            return;
+        }
+        const size_t slen = strlen(argv[i]) + 1;
+        char *newvar = (char *)calloc(slen, sizeof(char));
+        strncpy(newvar, argv[i], slen);
+
+        epicsMutexMustLock(thecaster.lock);
+        thecaster.extra_envs[thecaster.num_extra_envs - 1] = newvar;
+        thecaster.extra_envs[thecaster.num_extra_envs] = NULL;
```
simon-ess: I realised after the fact that we do not need this null-termination; the reason it is necessary for the `default_envs` variable is that there is no additional count of variables, so we use the null-termination to signify that we are done with the list. However, we do maintain a count, so we can avoid having the additional null termination.

client/castApp/src/caster.c
```diff
@@ -129,6 +129,16 @@ void casterShutdown(caster_t *self)
...
     epicsMutexUnlock(self->lock);
 
+    if(self->extra_envs) {
+        epicsMutexMustLock(self->lock);
+        int i;
+        for(i = 0; self->extra_envs[i]; i++) {
+            free(self->extra_envs[i]);
+        }
+        free(self->extra_envs);
+        epicsMutexUnlock(self->lock);
```
simon-ess: Do we need this extra lock/unlock in here? Can't we just move the `epicsMutexUnlock`from above to below this conditional block?

client/castApp/src/dbcb.c
```diff
@@ -58,14 +62,23 @@ static int pushEnv(caster_t *caster)
...
         if(val && val[0]!='\0')
-            ret = casterSendInfo(caster, 0, envs[i], val);
+            ret = casterSendInfo(caster, 0, default_envs[i], val);
         if(ret)
-            casterMsg(caster, "Error sending env %s", envs[i]);
+            casterMsg(caster, "Error sending env %s", default_envs[i]);
     }
 
+    if(caster->extra_envs) {
+        for (i = 0; !ret && caster->extra_envs[i]; i++) {
```
simon-ess: I think if one removes the NULL-termination of the `extra_envs` variable this can likely become simpler; we don't even need the `if (caster->extra_envs)` for example.

client/castApp/src/caster.c
tynanford: Yes we could do that. I thought there might be some benefit to be explicit about which parts need locking. But that's probably not useful

client/castApp/src/castinit.c
tynanford: yes this is true. Do you think there is any benefit with having the extra safety of being null terminated?

simon-ess: Not obviously to me, no. I think it risks confusion later on in case one of the redundant checks falls out of sync.

client/castApp/src/caster.c
mdavidsaver: This block looks a little odd to me.  Seeing locking is necessary around `free(self->extra_envs)`, I would expect it is also necessary around `if(self->extra_envs)` (the pointer can change).

Also, this whole block should be moved after the `epicsEventMustWait(self->shutdownEvent);`, by which point the worker thread will not access it.

fyi. The worker shutdown process has three parts. 

1. Set the `shutdown` flag
2. Write a byte to `wakeup` to ensure the worker will check the `shutdown` flag
3. Wait until `casterThread()` signals the `shutdownEvent` just prior to returning.

client/castApp/src/castinit.c
mdavidsaver: Having a `realloc()` in this loop isn't so efficient,  Also, having multiple lock/unlock operations in the loop seems like an unnecessary invitation for bugs.  This requires that code accessing `extra_envs[]` be aware that some elements may be `NULL` (and perhaps not just the last).

Personally, I see "coarse" locking as a safer starting position (aka. lock over a much as possible).  imo. Finer grained locking should be seen as an optimization when necessary, and with due consideration given to possible concurrent access.

Also, wrt. `var = realloc(var, ...);` keep in mind that `realloc()` does __not__ `free()` anything on error.  So using the same variable for both result and argument is an easy way to leak on error.  (recent GCC should warn about this)

If you want to optimize to avoid all of the `calloc()`s while locked, then I would suggest making two loops.  One without locking, which looks only at the argument list.  Then a second with locking, to modify the `caster_t*`.

eg. the following skeleton, which omits printing error messages

```c
static void addReccasterEnvVars(caster_t* self, int argc, char **argv) {
  int i;
  int ret = 0;
  char **newelems;
  if(argc<2)
    return; // one var required
  /* temporary newelems[] to hold individually allocated strings */
  if(!(newelems = calloc(argc-1, sizeof(*newelems))))
    ret = 1;
  for(i=0; i<argc-1 && !ret; i++) {
    if(!(newelems[i] = strdup(argv[i+1])))
      ret = 1;
  }
  if(!ret) {
    char **new_envs;
    epicsMutexMustLock(self->lock);
    if(self->shutdown) {
      /* shutdown in progress, silent no-op */
    } else if(self->state != casterStateInit) {
      /* Attempt to add after iocInit(), when we may be connected.
       * To fully support, would need to force reconnect or resend w/ updated envs list.
       */
       ret = 2;
    } else if(!(new_envs = realloc(self->extra_envs,
                                   sizeof(*new_envs) * (self->num_extra_envs + argc - 1)))) {
      ret = 1;
    } else {
        /* from this point, nothing can fail */
        self->extra_envs = new_envs;
        self->num_extra_envs += argc - 1;

        for(i=0; i<argc-1; i++) {
          new_envs[self->num_extra_envs + i] = newelems[i];
          newelems[i] = NULL; /* prevent early free below */
        }
    }
    epicsMutexUnlock(self->lock);
  }
  /* cleanup newelems[] on success or failure */
  for(i=0; i<argc-1 && !ret; i++) {
    free(newelems[i]);
  }
  free(newelems);
  if(ret) ... oops
}
```

Note that this skeleton ignores the possibility of duplicate variable names.

client/castApp/src/dbcb.c
mdavidsaver: Need some locking for accesses of and through `extra_envs[]`.

mdavidsaver:
fyi. https://github.com/epics-base/pvAccessCPP/pull/185

client/castApp/src/castinit.c
tynanford: Thanks @mdavidsaver I'll have a look more at your example. Just quick question:

> which omits printing error messages

Should I take this as meaning I shouldn't print error messages and just return instead?

client/castApp/src/caster.c
tynanford: resolved with https://github.com/ChannelFinder/recsync/pull/57/commits/7baa5f8d08a1386656d876fa2a81519f34ce5aad

client/castApp/src/castinit.c
mdavidsaver: > Should I take this as meaning I shouldn't print error messages and just return instead?

Not at all!  My "example" is long enough without wordy prints.

client/castApp/src/dbcb.c
tynanford: Resolved both comments w/ https://github.com/ChannelFinder/recsync/pull/57/commits/908052b7244a1ed8b477216a6f0cc3a41d5b3176

client/castApp/src/castinit.c
tynanford: Hi @mdavidsaver ,

I pushed a commit to try to address this and worked off your skeleton. I had some questions while writing it up:

In your skeleton I see a check on `!ret` for the cleanup of the temporary array. I think we don't want this check since we want to clean up that memory either way? I didn't include the check in the commit but wanted to run by you
```
  /* cleanup newelems[] on success or failure */
  for(i=0; i<argc-1 && !ret; i++) {
```

I added logic to check if someone tries to add an environment variable that is already in `self->extra_envs`. This could be extended to also check against the default_envs list too (and probably should) but that isn't in there right now. With this logic check, I had to add a second lock/unlock for accessing self->extra_envs. Also that brings up the question of should I just have a single lock given what you mentioned previously 

After adding the second lock, I thought about moving the checks on this->shutdown and this->current to the top and created a commit for that. I didn't end up going with it since I thought there might be good reason to keep those checks in the same lock/unlock block with where we edit `self->extra_envs`. But if you are curious you can view that idea here: https://github.com/tynanford/recsync/commit/f566ba228d54c42b946496e78dcbe79e563075b0

Appreciate the help

client/castApp/src/castinit.c
```diff
@@ -65,13 +67,115 @@ static void casthook(initHookState state)
...
+    }
+    for(i = 0, j = 0; i < argc - 1; i++) {
+        /* check for bad args which are set to NULL above */
+        if(!argv[i+1])
+            continue;
+        if((tmp_new_envs[j] = strdup(argv[i+1])) == NULL) {
+            ret = 1;
+        }
+        j++;
+    }
```
simon-ess: Very minor nitpick, but this for loop runs from 0 to `argc - 1` (with index `i+1`) while the one above runs from 1 to `argc` (with index `i`). It might be nicer to harmonise this...

simon-ess:
Is there any progress on getting this merged in? This is functionality that we could happily use, but it seems to currently be sitting in limbo.

tynanford:
Hi Simon, I will be at the code-a-thon at Diamond in a few weeks and was hoping to discuss this with yourself and @mdavidsaver or others if you guys will not be there. At this point I am waiting on a code review

client/castApp/src/castinit.c
tynanford: agreed. I made a note of this and add it in with the remaining cleanup most likely that is needed

tynanford: updated these loops to match

tynanford:
@mdavidsaver ready for a review

client/castApp/src/castinit.c
```diff
@@ -65,13 +67,115 @@ static void casthook(initHookState state)
...
+*/
+static void addReccasterEnvVars(caster_t* self, int argc, char **argv)
+{
+    int i, j;
+    int ret = 0;
+    if(argc < 2) {
+        errlogSevPrintf(errlogMinor, "At least one argument expected for addReccasterEnvVars\n");
+        return;
+    }
+    /* sanitize input - check for dups and empty args. First arg in argv is function name so skip that */
```
mdavidsaver: fyi. Another way to skip the function name would be

```c
argv++;
argc--;
```

client/castApp/src/castinit.c
```diff
@@ -65,13 +67,115 @@ static void casthook(initHookState state)
...
+            errlogSevPrintf(errlogMinor, "Arg is NULL for addReccasterEnvVars\n");
+            continue;
+        }
+        else if(argv[i][0] == '\0') {
+            argCount--;
+            errlogSevPrintf(errlogMinor, "Arg is empty for addReccasterEnvVars\n");
+            argv[i] = NULL;
+            continue;
+        }
+        /* check if dup in self->extra_envs. doesn't check if arg is in default_envs right now */
```
mdavidsaver: Are you concerned about duplicates in the argument list as well?

Maybe a simpler approach would be to

1. allocate space for `num_extra_envs+argc` as `new_extra_envs[]`
2. copy in existing `extra_envs[]`
3. for each argv[i], look for duplicates in `new_extra_envs[]`, then either append to skip.

This would avoid the need to look multiple times.

tynanford:
Thanks for the review @mdavidsaver . Latest commit updates to use the argv++; argc--; to skip function name and uses the validation logic you mentioned to keep things in one loop. Could you take another look?

client/castApp/src/castinit.c
```diff
@@ -65,13 +67,133 @@ static void casthook(initHookState state)
...
+            }
+            else if(argv[i][0] == '\0') {
+                bad_arg_count++;
+                errlogSevPrintf(errlogMinor, "Arg is empty for addReccasterEnvVars\n");
+                continue;
+            }
+            found_dup = 0;
+            /* check if dup in self->extra_envs. doesn't check if arg is in default_envs right now */
+            for(k = 0; k < num_new_extra_envs && new_extra_envs[k]; k++) {
+                if(strcmp(argv[i], new_extra_envs[k]) == 0) {
```
mdavidsaver: Can `new_extra_envs[k]==NULL` here?  I think that `strcmp()` and other `<string.h>` functions are undefined when passed `NULL`.  So I think you will need to separately track of both total capacity, and current size, for the new array.

client/castApp/src/castinit.c
```diff
@@ -65,13 +67,133 @@ static void casthook(initHookState state)
...
+                break;
+            }
+            /* this is a valid arg and we have added the new env var to our array, increment new_extra_envs index */
+            j++;
+        }
+    }
+    /* if we have no allocation issues and have at least one new env var that is valid, add to self->extra_envs */
+    if(!ret && bad_arg_count != argc) {
+        char ** new_envs;
+        new_envs = realloc(self->extra_envs, sizeof(* new_envs) * (self->num_extra_envs + (argc-bad_arg_count)));
```
mdavidsaver: Why not replace `self->extra_envs` with `new_extra_envs` here instead of another alloc+copy?  In the common case (no duplicates) they will be identical, and in the uncommon case only a few bytes are wasted.

tynanford:
@mdavidsaver how does that look?

client/castApp/src/castinit.c
```diff
@@ -65,13 +67,124 @@ static void casthook(initHookState state)
...
+                errlogSevPrintf(errlogMinor, "strdup error for copying %s to new_extra_envs[%d] from addReccasterEnvVars\n", new_extra_envs[num_new_extra_envs]);
+                ret = 1;
+                break;
+            }
+            /* this is a valid arg and we have added the new env var to our array, increment new_extra_envs count */
+            num_new_extra_envs++;
+        }
+    }
+    /* if we have no allocation issues and have at least one new env var that is valid, add to self->extra_envs */
+    if(!ret && num_new_extra_envs - self->num_extra_envs != 0) {
```
simon-ess: Any particular reason you check the difference and not just num_new_extra_envs != self->num_extra_envs?

(or even num_new_extra_envs > self->num_extra_envs as a bit of a sanity check)

tynanford: nope no reason besides not thinking right :)

client/castApp/src/castinit.c
```diff
@@ -65,13 +67,124 @@ static void casthook(initHookState state)
...
+                errlogSevPrintf(errlogMinor, "Arg is NULL for addReccasterEnvVars\n");
+                continue;
+            }
+            else if(argv[i][0] == '\0') {
+                errlogSevPrintf(errlogMinor, "Arg is empty for addReccasterEnvVars\n");
+                continue;
+            }
+            found_dup = 0;
+            /* check if dup in self->extra_envs. doesn't check if arg is in default_envs right now */
+            for(j = 0; j < num_new_extra_envs && new_extra_envs[j]; j++) {
```
simon-ess: I think that this could be split up into a test within the for loop. If somehow or other a `NULL` sneaks its way into `self->extra_envs`, then it would no longer check anything past that point.

(I think that the code is also a bit easier to follow with a separate condition in the loop instead of including it as the loop condition)

client/castApp/src/testAddEnvVars.c
```diff
@@ -0,0 +1,145 @@
...
+{
+    int i;
+    lock = epicsMutexMustCreate();
+
+    caster_t caster;
+    casterInit(&caster);
+
+    epicsMutexUnlock(lock);
+
+    epicsMutexMustLock(lock);
```
simon-ess: Why is there a lock in this test?

client/castApp/src/testAddEnvVars.c
```diff
@@ -0,0 +1,145 @@
...
+    testOk1(caster.num_extra_envs==expectedNumExtraEnvs);
+    addReccasterEnvVars(&caster, argc, argvlist);
+    testOk1(caster.num_extra_envs==expectedNumExtraEnvs);
+
+    testDiag("Testing addReccasterEnvVars with NULL argument");
+    argvlist[1] = NULL;
+    argc = 2;
+    testOk1(caster.num_extra_envs==expectedNumExtraEnvs);
+    addReccasterEnvVars(&caster, argc, argvlist);
+    testOk1(caster.num_extra_envs==expectedNumExtraEnvs);
```
simon-ess: I see that in general you are testing repeated calls to `addReccasterEnvVars` and so they need to be in a common test, but I think that these simpler ones should be a separate test function for clarity's sake.

client/castApp/src/testAddEnvVars.c
```diff
@@ -0,0 +1,145 @@
...
+    for(i=0; i < expectedNumExtraEnvs; i++) {
+        testOk1(strcmp(caster.extra_envs[i], expectedExtraEnvs[i]) == 0);
+    }
+
+    epicsMutexUnlock(lock);
+
+    epicsEventId sd;
+    sd = caster.shutdownEvent;
+    epicsEventSignal(sd);
```
simon-ess: This looks to me to be an entirely separate test about the startup and shutdown mechanism. Is it needed here?

tynanford: we determined this is needed to get casterShutdown to complete since there is epicsEventMustWait(self->shutdownEvent); inside

client/castApp/src/testAddEnvVars.c
tynanford: discussed with Simon and determined this is not needed

client/castApp/src/testAddEnvVars.c
```diff
@@ -0,0 +1,162 @@
...
+    testOk1(caster.num_extra_envs==expectedNumExtraEnvs);
+    addReccasterEnvVars(&caster, argc, argvlist);
+    testOk1(caster.num_extra_envs==expectedNumExtraEnvs);
+    for(i=0; i < expectedNumExtraEnvs; i++) {
+        testOk1(strcmp(caster.extra_envs[i], expectedExtraEnvs[i]) == 0);
+    }
+
+    epicsEventId sd;
+    sd = caster.shutdownEvent;
+    epicsEventSignal(sd);
```
simon-ess: Is there any reason you can't just do `epicsEventSignal(caster.shutdownEvent)`?

tynanford: yes that is cleaner, updated. thanks!

PR #58, opened 2022-05-17
Fix CI with rhel7 in docker addition

tynanford:


tynanford:
I think I see an issue here with what I did. Probably didn't clone with the submodules which is why I had to re-add it. I can fix that

tynanford:
closed in favor of https://github.com/ChannelFinder/recsync/pull/59

PR #59, opened 2022-05-17
Update CI

tynanford:


tynanford:
@mdavidsaver this is ready to merge from my point of view. What do you think?

I basically followed what you did here: https://github.com/mdavidsaver/devlib2/commit/26574ffb6890bce6376e66543f0dd2760e99ff5f

shroffk (approved):

PR #64, opened 2022-10-31
Updates per discussion at code-a-thon

tynanford:
- move iocsh register code to castinit.c and place in reccasterRegistrar
- make addReccasterEnvVars take argv instead of single string
- Add PVAS_SERVER_PORT and RSRV_SERVER_PORT to default variables to send


tynanford:
didn't mean to create new PR... closing

PR #76, opened 2024-01-31
Fix return check

FreddieAkeroyd:
Fix check - `shCreateSocket` returns `INVALID_SOCKET` on error which is usually `-1`


mdavidsaver:
Thanks Freddie.

Issue #74, opened 2023-09-15
Issue deploying mulitple recceiver instances


tynanford:
When deploying multiple recceivers pushing to one channel finder instance, there is an issue with the clean_service method where it will set all pvStatus properties on all PVs in CF to "Inactive" on recceiver startup. This works fine if you have one recceiver but once there are more, as you start each recceiver then it will mark all the PVs belonging to the other recceiver instances "Inactive"

Setting the cleanOnStart/cleanOnStop preference setting stops the clean_service method from getting called but I think you then lose some functionality. Updating the CF query in clean_service with a preference setting to allow you to limit the scope of the clean could be useful but not sure how to limit it. Query on iocid since usually it's a per subnet basis for each instance? 

https://github.com/ChannelFinder/recsync/blob/master/server/recceiver/cfstore.py#L284

shroffk:
I agree.. I think that might be the issue.

tynanford:
fixed via https://github.com/ChannelFinder/recsync/pull/78 , thanks @jacomago !

PR #78, opened 2024-03-06
Add recceiver id, which is used by cleanOnStart and cleanOnStop

jacomago:


server/recceiver/cfstore.py
```diff
@@ -521,6 +512,15 @@ def __updateCF__(proc, pvInfoByName, delrec, hostName, iocName, iocid, owner, io
...
 
+def create_properties(owner, iocTime, channels_dict, iocs, recceiver_id, ch):
+    return [
+                        {u'name': 'hostName', u'owner': owner, u'value': iocs[channels_dict[ch[u'name']][-1]]["hostname"]},
+                        {u'name': 'iocName', u'owner': owner, u'value': iocs[channels_dict[ch[u'name']][-1]]["iocname"]},
+                        {u'name': 'iocid', u'owner': owner, u'value': channels_dict[ch[u'name']][-1]},
+                        {u'name': 'pvStatus', u'owner': owner, u'value': 'Active'},
+                        {u'name': 'time', u'owner': owner, u'value': iocTime},
+                        {u'name': RECCEIVER_ID_KEY, u'owner': owner, u'value': recceiver_id}]
+
```
simon-ess: While I do really like these cleanups, it might have been better to put them in a separate commit to make it clear what is cleanup, and what is the functional change.

server/recceiver/cfstore.py
```diff
@@ -467,7 +457,8 @@ def __updateCF__(proc, pvInfoByName, delrec, hostName, iocName, iocid, owner, io
                      {u'name': 'iocName', u'owner': owner, u'value': iocName},
                      {u'name': 'iocid', u'owner': owner, u'value': iocid},
                      {u'name': 'pvStatus', u'owner': owner, u'value': "Active"},
-                     {u'name': 'time', u'owner': owner, u'value': iocTime}]
+                     {u'name': 'time', u'owner': owner, u'value': iocTime},
+                     {u'name': RECCEIVER_ID_KEY, u'owner': owner, u'value': recceiver_id}]
```
simon-ess: Given your new function `create_properties` has a lot of overlap with this, isn't it possible to use it here as well?

server/recceiver/cfstore.py
```diff
@@ -224,7 +227,7 @@ def _commitWithThread(self, TR):
         pvInfoByName = {}
         for rid, (info) in pvInfo.items():
             if info["pvName"] in pvInfoByName:
-                _log.warn("Commit contains multiple records with PV name: %s (%s)", pv, iocid)
+                _log.warn("Commit contains multiple records with PV name: %s (%s)", info["pvName"], iocid)
```
simon-ess, line 231: Good catch...

simon-ess (commented):
Broadly speaking this seems like a good change. My one main concern is that leaving it up to the user to configure the RECCEIVER_ID seems potentially problematic. What happens if multiple recceivers share a key? Is there no way to use something like `hostname` or something similar to specify the ID?

server/recceiver/cfstore.py
jacomago: done

server/recceiver/cfstore.py
jacomago: done

tynanford:
This is a good idea and would address https://github.com/ChannelFinder/recsync/issues/74 nicely. I agree with @simon-ess that using something like the hostname would help reduce manual config and user errors. What do you think @jacomago ?

Could you update cf.conf and demo.conf with the new config option `RECCEIVER_ID` so it's more visible to new users?

server/recceiver/cfstore.py
```diff
@@ -502,20 +507,22 @@ def __updateCF__(proc, pvInfoByName, delrec, hostName, iocName, iocid, owner, io
     if proc.cancelled:
         raise defer.CancelledError()
 
-def create_properties(owner, iocTime, hostName, iocName, iocid):
+def create_properties(owner, iocTime, recceiver_id, hostName, iocName, iocid):
```
jeonghanlee: I think, we should use the same convention here between `receiver_id` and `iocid`. Please select it either convention to keep the similar style, even if we have multiple convention. I don't want to see more conventions in a code.

simon-ess: I think that's good advice; there are already two conventions in that line (`hostName` vs. `iocid`), adding a third is probably overkill :smile:

jacomago: Updated

jacomago:
> This is a good idea and would address #74 nicely. I agree with @simon-ess that using something like the hostname would help reduce manual config and user errors. What do you think @jacomago ?
> 
> Could you update cf.conf and demo.conf with the new config option `RECCEIVER_ID` so it's more visible to new users?

Set the default value to be the hostname and updated the conf files.

simon-ess:
LGTM. It would be nice in theory to fix the failing pipelines, but those are pretty clearly a separate task that should be handled otherwise (in particular the Centos7 ones; I haven't seen the other failures before)

simon-ess (approved):

PR #80, opened 2024-04-03
Fix ci

jacomago:
Update the client ci scripts to align with the recsync client submodule ci.
Make sure all the builds pass

### DIFF
--- a/.github/workflows/ci-scripts-build.yml
+++ b/.github/workflows/ci-scripts-build.yml
@@ -21,223 +21,138 @@ env:
     EPICS_TEST_IMPRECISE_TIMING: YES
 
 jobs:
-  build-client:
-    name: ${{ matrix.name }}
-    runs-on: ${{ matrix.os }}
-    # Set environment variables from matrix parameters
-    env:
-      BASE: ${{ matrix.base }}
-      CMP: ${{ matrix.cmp }}
-      BCFG: ${{ matrix.configuration }}
-      CI_CROSS_TARGETS: ${{ matrix.cross }}
-      EXTRA: ${{ matrix.extra }}
-      TEST: ${{ matrix.test }}
-    strategy:
-      fail-fast: false
-      matrix:
-        include:
-          - name: Linux-Mingw-7.0
-            os: ubuntu-latest
-            cmp: gcc
-            configuration: default
-            base: "7.0"
-            cross: "windows-x64-mingw"
 
-          - name: Linux-static-7.0
-            os: ubuntu-latest
-            cmp: gcc
-            configuration: static
-            base: "7.0"
-
-          - name: Linux-3.15
-            os: ubuntu-latest
-            cmp: gcc
-            configuration: default
-            base: "3.15"
-
-          - name: Linux-3.14
-            os: ubuntu-latest
-            cmp: gcc
-            configuration: default
-            base: "3.14"
-
-          - name: Linux-c++11-7.0
-            os: ubuntu-latest
-            cmp: gcc
-            configuration: static
-            base: "7.0"
-            extra: "CMD_CXXFLAGS=-std=c++11"
-
-          - name: Linux-clang-7.0
-            os: ubuntu-latest
-            cmp: clang
-            configuration: default
-            base: "7.0"
-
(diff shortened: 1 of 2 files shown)


jeonghanlee:
@jacomago could you describe your merge request in more specific way in your comment? This is not the personal repository, and not the ESS specific repository. All other facilities should know what you are trying to do through this merge.

We don't have the template, but you can find another pull request as your example which Michael did in this repository.

jacomago:
@mdavidsaver Do you know why the macos12 ones are flaky? They seem to pass on reruns...

mdavidsaver:
> @mdavidsaver Do you know why the macos12 ones are flaky? They seem to pass on reruns...

I have probably encoded some Linux specific assumption into that test.  I don't immediately see what though.  100 second really should be long enough for even the slowest GHA runner.  It might be helpful for someone with a mac to try running this test locally.

jacomago:
@mdavidsaver Locally I'm getting a failure on 100 seconds but not on 110, and flaky inbetween. Shall I just bump the number? I don't know if this is to do with cpu seconds vs real time seconds...

simon-ess:
Regarding the mac os test flakiness, this seems to have been a long-standing issue. Even going back to 1.0 we see that the socket timeout test for mac os just... fails some of the time. It might be better to make that test be skipped on mac os...

client/castApp/src/testsock.c
```diff
@@ -100,10 +102,16 @@ static void testWakeup(void)
...
+        testPlan(18);
+    } else {
+        testPlan(14);
+    }
     osiSockAttach();;
     testUDP();
-    testWakeup();
+    if (!getenv("SKIP_TEST_WAKEUP")) {
+        testWakeup();
+    }
```
simon-ess: I think that a better option is to use `testSkip`, something akin to
```
    testPlan(18);
    osiSockAttach();;
    testUDP();
    if (!getenv("SKIP_TEST_WAKEUP")) {
        testWakeup();
    } else {
        testSkip(4);
    }
```

simon-ess: Even better (IMO) would be something like
```c
#ifdef __MACH__
    testSkip(4);
#else
    // everything else
#endif
```
or whatever the correct pre-processor directive is.

simon-ess:
Out of curiosity, why split up the linux/mac/windows tests? Those are doing the exact same things each, aren't they?

ralphlange:
Mac: different compiler, no static builds, no cross builds.
Windows: different compilers, no cross builds, build git from sources first.

On GitHub Actions, availability of runners may differ a lot by OS. You get a better overall timing by parallelizing the jobs that usually wait longer. ("Start waiting _now_.")

simon-ess:
My question was more about the matrix used for each case; the steps are the same, but they simply use a different matrix. However, looking a bit more carefully it seems that it isn't an easy fix; it's sort of like several "block diagonal" matrices (to abuse mathematical terms) and it isn't really easy to split that up without some possible yaml anchoring or something similar. Maybe this is best and cleanest.

.github/workflows/ci-scripts-build.yml
```diff
@@ -21,223 +21,138 @@ env:
...
-      VV: "1"
     strategy:
       fail-fast: false
       matrix:
-        include:
-          - name: Linux centos 7
-            image: centos:7
-            cmp: gcc
-            configuration: default
-            base: "7.0"
```
simon-ess, line 186: I see that we are removing centos7 support from CI?

ralphlange: They have.

The only way to run CentOS7 is in self-maintained containers.

simon-ess: I thought that might be the case, but it was not mentioned in the pull request.

jacomago:
@mdavidsaver or @ralphlange  review?

ralphlange:
Looks good to me.

(Note that - in other repos - I started adding CentOS-8 and Rocky-9 docker builds instead of the CentOS-7 that won't work on GHA anymore. No hurry - just saying.)

ralphlange (approved):
LGTM.

PR #81, opened 2024-04-12
Fix cleaning > 10000 channels

jacomago:
Elastic/ChannelFinder has a setting ES_QUERY_SIZE. So by default queries only return up to ES_QUERY_SIZE channels, which by default is 10000.

When the clean runs it fetches using this api all channels where the pvStatus is Active. But if there are more than 10000 this won't work due to the ES_QUERY_SIZE.

Solution here is to keep setting channels to Inactive and querying for more channels until there are no channels set to Active.

### DIFF
--- a/server/recceiver/cfstore.py
+++ b/server/recceiver/cfstore.py
@@ -282,17 +282,14 @@ class CFProcessor(service.Service):
         while 1:
             try:
                 _log.info("CF Clean Started")
-                channels = self.client.findByArgs(prepareFindArgs(self.conf, [('pvStatus', 'Active'), (RECCEIVERID_KEY, recceiverid)]))
+                channels = self.get_active_channels(recceiverid)
                 if channels is not None:
-                    new_channels = []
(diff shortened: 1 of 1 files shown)


jacomago:
Will do more testing next week to check this works as expected.

tynanford:
I ran into this issue and tried the PR. In most cases it seemed like 20K PVs got cleaned on shutdown out of 38K PVs. But this is using docker and docker-compose to run recceiver and that looks like the reason it doesn't fully work? Using the --timeout flag like `docker compose up --timeout 600` let the clean process finish and all the PVs were marked Inactive.

Is there a reason to hard-code 10000 in the `findByArgs` call? I think it would be better to either have the user set `findSizeLimit` in their config or default to the ES limit

tynanford (approved):
LGTM! Added @shroffk as reviewer too

shroffk:
LGTM

However, I think we should work on a better API for these kinds of operations

PR #82, opened 2024-04-16
Make a minimal build and install ci pipeline

jacomago:
Add requests and channelfinder client as dependancies.
Add a minimal install for server.


tynanford (approved):
LGTM!

PR #83, opened 2024-04-18
Add more logging

jacomago:
Add some extra logs in various places

Use twisted logging not python logging

  The python logger is not passing to the twisted
  logger by default or with the Log2Twisted setup.
  
  From the twisted documentation
  https://docs.twisted.org/en/twisted-18.7.0/core/howto/logger.html#compatibility-with-standard-library-logging
  It's suggested not to do this anyway.
  
  This commit moves all the logs to the twisted
  logger twisted.logger.Logger.

### DIFF
--- a/server/recceiver/announce.py
+++ b/server/recceiver/announce.py
@@ -4,9 +4,9 @@ import sys
 import struct
 
 from twisted.internet import protocol, reactor
-import logging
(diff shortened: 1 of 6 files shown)


tynanford:
makes sense to me and some nice cleanup

tynanford (approved):

PR #85, opened 2024-04-18
update readfp to read_file for python 3.12

tynanford:
`readfp` was depcrecated in python 3.2 and removed in python 3.12

The replacement is read_file which was introduced in python 3.2: https://docs.python.org/3/library/configparser.html#configparser.ConfigParser.read_file

This keeps things backwards compatible and but at some point we should create the last python2/3 release and then go forward with python 3 only?

### DIFF
--- a/server/recceiver/processors.py
+++ b/server/recceiver/processors.py
@@ -69,7 +69,10 @@ class ProcessorController(service.MultiService):
         read = parser.read(map(expanduser, self.paths))
 
         if cfile:
-            parser.readfp(open(cfile,'r'))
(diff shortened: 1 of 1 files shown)


jacomago (changes requested):
The borderline of the change is at 3.2, so better to keep to that I think.

https://docs.python.org/3.11/library/configparser.html#configparser.ConfigParser.read_file

tynanford:
sounds good to me, updated the check

jacomago (approved):

PR #88, opened 2024-05-08
Sets the support level to python 3.6 or above

jacomago:
Removes from the server support for python 3.2 and below
Sets the suggested python version to 3.6 or above

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ RecCeiver Usage
 
 The RecCeiver server in the `server/` directory is a
 Python script using the [Twisted][twisted] networking
-library.  It requires Python 2.7 and Twisted >= 12.0.
+library.  It requires Python 3.6 or above and Twisted >= 12.0.
 
 [twisted]: http://twistedmatrix.com/
 
--- a/server/recceiver/application.py
+++ b/server/recceiver/application.py
@@ -25,11 +25,8 @@ class Log2Twisted(logging.StreamHandler):
     def __init__(self):
         super(Log2Twisted,self).__init__(stream=self)
         # The Twisted log publisher adds a newline, so strip the newline added by the Python log handler.
-        if sys.version_info < (3,2):
-            self.write = lambda *args, **kwargs: log.msg(*[ str(a).strip() for a in args ], **kwargs)
-        else:
-            self.terminator = ""  #  the 'terminator' attribute was added to StreamHandler in Python v3.2
-            self.write = log.msg
+        self.terminator = ""
+        self.write = log.msg
     def flush(self):
         pass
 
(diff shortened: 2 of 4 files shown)


jacomago:
As support is being removed more and more for lower python versions. I thought it would make sense to make this change. I expect a lot of sites use a docker image for recceiver anyway, but I will leave this PR unmerged for some time.

jeonghanlee:
@jacomago could you collect all user OS requirements? That changes are okay for the ALS-U, since we officially support Debian 12 and Rocky 9.2. Both OSs have the Python 3.11. However, what about other facilities? 

I am still thinking these changes are a bit early.

Why not we keep the old codes as it is, and add new version code together? Or do you have any specific reasons regarding security concerns?

ralphlange:
I agree with Han.

Don't underestimate user annoyance when doing your cost benefit analysis. 

RHEL is widely used and pretty conservative.  To me, the standard Python version on RHEL 8 would look like a good marker for what people might want to use.

tynanford:
RHEL 8 ships with python 3.6.8 and I agree with Ralph to at least support that OS. Looking naively at the PR, it just drops python 2 support? What changes preclude python 3 below 3.7?

RHEL 7 is still around til the end of June and ships with python 2.7 but maybe those users should be focusing more on updating their OS than upgrading recceiver to the latest version in github ;) RHEL 8 is dropping support for python 2.7 at the same time - https://access.redhat.com/solutions/4455511

So dropping python 2 soon would be reasonable to me.

jeonghanlee:
I also agree with Tynan, we should prepare to drop python 2 soon.

shroffk:
To sum up

- We should drop python 2 support.
- @jacomago are there any changes needed to keep the minimum support to 3.6. Like @tynanford mentioned, I think there are some changes which need at least 3.2 and the rest are simply getting rid of the switch added for supporting both version 2 and 3.
- since a large number on EPICS institutions use red hat. We can use that to decide the minimum versions we should support. Currently 3.6 seems like a good target (unless we have a strong justification for features in a newer release).

jacomago:
Perfect. Yes, as @tynanford this actually removes support for 3.2 and below. I've updated the PR description. I'll make a follow up PR to add building down to 3.6.

jeonghanlee:
@jacomago 3.2 and below?

So do we support 3.3.X after this merge? Or 3.6.X?

Can we specify the future support version more clearly?

jacomago:
@jeonghanlee I think its best to say this PR keeps the code working for python >=3.3. But we keep support specified in the pyproject.toml which is set to >=3.6. This makes clear we won't fix bugs for running on python <=3.5.

tynanford:
I think technically it is >=3.2 like @shroffk mentioned. It would be good to update the README with this PR as well -  https://github.com/ChannelFinder/recsync?tab=readme-ov-file#recceiver-usage

ralphlange:
Might be nitpicking...
Could you update the commit message to better reflect the change? (Could scare people...)

jacomago:
> Might be nitpicking... Could you update the commit message to better reflect the change? (Could scare people...)

Updated the commit and the PR descriptions.

shroffk (approved):

PR #89, opened 2024-05-08
Add the timezone to the time output

jacomago:
Adds the timezone to the time string on ChannelFinder updates

For example,
timezone = False time produces:
2024-05-08 12:55:18.722180
timezone = True (Berlin Time) time produces:
2024-05-08 12:55:18.722180+02:00


tynanford (approved):
LGTM

georgweiss (approved):

Issue #37, opened 2019-01-02
twistd errors with "reactor already installed" when dropin.cache was not created


shengb:
When running the twistd, if dropin.cache was not created yet, it reports
twistd: The specified reactor cannot be used, failed with error: reactor already installed.

A similar issue with the ticket on:
https://twistedmatrix.com/trac/ticket/3785

DanielALS:
Is there a solution to this ? 
I had to wipe out the dropin.cache to get things working again and hit this problem.

DanielALS:
UPDATE: Wrote too soon. This solution did not help me.

OK, found the solution to re-generate the cache:

https://twistedmatrix.com/documents/current/core/howto/plugin.html

    from twisted.plugin import IPlugin, getPlugins
    list(getPlugins(IPlugin))

tynanford:
This was fixed via https://github.com/ChannelFinder/recsync/pull/93

PR #93, opened 2024-08-07
Defer importing `twisted.internet.reactor`, fixes "reactor already installed"

minijackson:
While running RecCeiver with 24.3.0, I found the same issue as #37:

    The specified reactor cannot be used, failed with error: reactor already installed.

but initializing a `dropin.cache` did not work out for me. After some research, I found scrapy's asyncio documentation:

https://docs.scrapy.org/en/latest/topics/asyncio.html#handling-a-pre-installed-reactor

Turns out importing `twisted.internet.reactor` initializes a default reactor, which makes it impossible to change the reactor in the command-line (`-r` option).

So this PR defers importing reactor, which makes RecCeiver work for me.

This PR was only crudely tested, by seeing that the application doesn't crash.

### DIFF
--- a/server/recceiver/announce.py
+++ b/server/recceiver/announce.py
@@ -3,7 +3,7 @@
 import sys
 import struct
 
-from twisted.internet import protocol, reactor
+from twisted.internet import protocol
 from twisted.logger import Logger
 
 _log = Logger(__name__)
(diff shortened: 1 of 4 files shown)


tynanford:
Nice! This PR works for me and avoids the old hacks we had to avoid the "reactor already installed" issue. Tested with twisted 16.4.0 and 22.4.0 and confirmed CF is populated as normal

shroffk (approved):

tynanford (approved):

PR #94, opened 2024-08-07
Remove min function call from debug statement in the clean service function

tynanford:
@jacomago I ran into this log statement which was causing a runtime error when `findSizeLimit` wasn't defined in my configuration file.

First I was going to change this to `self.conf.get` but then it seemed like this isn't necessary here? Since `get_active_channels` is already called and it's returned value is the `channels` arg in `clean_channels`?


jacomago (approved):
LGTM

PR #98, opened 2025-01-10
Add the ruff linter and formatter to recceiver

jacomago:
Adds the ruff linter and formatter https://docs.astral.sh/ruff/

Checks for problematic code and auto formats code.

Removes unused code


server/recceiver/test_cfstore.py
```diff
@@ -29,7 +29,7 @@ def test_3_iocs(self):
             6: ("ch6", "stringin"),
             7: ("ch7", "longout"),
         }
-        deferred = yield self.cfstore.commit(TR1)
+        _ = yield self.cfstore.commit(TR1)
```
simon-ess: I'm a bit surprised by this. Wouldn't it be better to just do
```py
yield self.cfstore.commit(TR1)
```
?

test-client.py
```diff
@@ -2,12 +2,12 @@
...
 
-import sys, socket
+import socket
 
 S = socket.socket(socket.AF_INET, socket.SOCK_DGRAM, 0)
 
 S.bind(('',5049))
 
-print S, S.fileno()
+print(S, S.fileno())
```
simon-ess: Wait, is this python2 or python3 code? Were these tests being run?

jacomago: I'm guessing python2 code. I think this is more code to create a quick and dirty python reccaster. Could just remove?

server/recceiver/test_cfstore.py
jacomago: Yes. Probably. Tempted to just remove this test file entirely as it is sooo out of date and incompatible with other cfstore changes.

server/recceiver/application.py
```diff
@@ -79,68 +81,75 @@ def privilegedStartService(self):
...
+        self.tcp = self.reactor.listenTCP(
+            self.port, self.tcpFactory, interface=self.bind
+        )
         try:
             self.tcp.startListening()
         except CannotListenError:
-            pass # older Twisted required this.
-                 # newer Twisted errors. sigh...
+            pass  # older Twisted required this.
+            # newer Twisted errors. sigh...
```
simon-ess: This comment is weirdly aligned.

server/recceiver/cfstore.py
```diff
@@ -187,94 +204,138 @@ def _commitWithThread(self, TR):
...
 
         """The unique identifier for a particular IOC"""
         iocid = host + ":" + str(port)
 
         pvInfo = {}
         for rid, (rname, rtype) in TR.addrec.items():
             pvInfo[rid] = {"pvName": rname}
-            if (self.conf.get('recordType', 'default' == 'on')):
-                pvInfo[rid]['recordType'] = rtype
+            if self.conf.get("recordType", "default" == "on"):
```
simon-ess: This  looks... weird. Should this be instead
```py
if self.conf.get("recordType", default="on"):
```
Otherwise, this is just equivalent to
```py
if self.conf.get("recordType", False):
```
???

server/recceiver/cfstore.py
```diff
@@ -187,94 +204,138 @@ def _commitWithThread(self, TR):
...
+                "channelcount": 0,
+            }
         if not TR.connected:
             delrec.extend(self.channel_dict.keys())
         for pv in pvInfoByName.keys():
             self.channel_dict[pv].append(iocid)  # add iocname to pvName in dict
             self.iocs[iocid]["channelcount"] += 1
             """In case, alias exists"""
-            if (self.conf.get('alias', 'default' == 'on')):
+            if self.conf.get("alias", "default" == "on"):
```
simon-ess: Here too. `"default"` is never equal to `"on"`.

server/recceiver/cfstore.py
```diff
@@ -187,94 +204,138 @@ def _commitWithThread(self, TR):
...
                 if pv in pvInfoByName and "aliases" in pvInfoByName[pv]:
                     for a in pvInfoByName[pv]["aliases"]:
-                        self.channel_dict[a].append(iocid)  # add iocname to pvName in dict
+                        self.channel_dict[a].append(
+                            iocid
+                        )  # add iocname to pvName in dict
```
simon-ess, line 308: This comment is both a bit superfluous and slightly incorrect.

server/recceiver/cfstore.py
```diff
@@ -187,94 +204,138 @@ def _commitWithThread(self, TR):
...
                         self.iocs[iocid]["channelcount"] += 1
         for pv in delrec:
             if iocid in self.channel_dict[pv]:
                 self.remove_channel(pv, iocid)
                 """In case, alias exists"""
-                if (self.conf.get('alias', 'default' == 'on')):
+                if self.conf.get("alias", "default" == "on"):
```
simon-ess: Another one.

server/recceiver/cfstore.py
```diff
@@ -356,96 +441,216 @@ def __updateCF__(proc, pvInfoByName, delrec, hostName, iocName, iocIP, iocid, ow
...
+            ):  # case: empty commit/del, remove all reference to ioc
+                if ch["name"] in channels_dict:
+                    ch["owner"] = iocs[channels_dict[ch["name"]][-1]]["owner"]
+                    ch["properties"] = __merge_property_lists(
+                        ch_create_properties(
+                            owner, iocTime, recceiverid, channels_dict, iocs, ch
+                        ),
+                        ch["properties"],
+                    )
+                    if conf.get("recordType", "default") == "on":
```
simon-ess: Ooooooh maybe this is what it should have been in those earlier examples.

server/recceiver/cfstore.py
```diff
@@ -567,26 +803,42 @@ def getCurrentTime(timezone=False):
...
             _log.error("ChannelFinder update failed: {s}".format(s=e))
             retry_seconds = min(60, sleep)
-            _log.info("ChannelFinder update retry in {retry_seconds} seconds".format(retry_seconds=retry_seconds))
-            #_log.debug(str(channels_dict))
+            _log.info(
+                "ChannelFinder update retry in {retry_seconds} seconds".format(
+                    retry_seconds=retry_seconds
+                )
+            )
+            # _log.debug(str(channels_dict))
```
simon-ess: Commented code

server/recceiver/recast.py
```diff
@@ -133,25 +137,29 @@ def recvClientGreeting(self, body):
 
     # 0x0002
     def recvPong(self, body):
-        nonce, = _ping.unpack(body[:_ping.size])
```
simon-ess, line 136: This looks odd to me, why is it not just
```py
nonce = _ping.unpack(...)
```
Or is the point that it is a list with a single element? I guess then
```py
nonce, *_ = _ping.unpack(...)
```
but maybe this doesn't matter...

jacomago: I think unpack returns a tuple (can actually see in the docs)

server/recceiver/cfstore.py
jacomago: Fixed by swapping to just if self.conf.get("recordType")

server/recceiver/recast.py
anderslindho: I though the convention was to prefer `foo,` over `(foo,)`

anderslindho (approved):
LGTM

PR #100, opened 2025-02-05
Add the infotags example back to cf.conf and demo.conf

### LINKED ISSUE #99: infotags removed from cf.conf file
Hello,

I was working to upgrade from Channel Finder 4.7.1 to 4.7.2 and noticed that the following was removed from cf.conf and demo.conf.
```
[cf]
# a space-separated list of infotags to set as CF Properties
#infotags = archive foo bar blah
```

Upon testing 4.7.2 without infotags declared, Channel Finder did not pick up any of those tags from PVs. When I added the lines back, uncommented, and restarted, my tags came through as they did in 4.7.1.

Should infotags be added to the config file or is there a new way to declare these?

Thank you,
Connor

jacomago:
Fixes #99


server/cf.conf
```diff
@@ -36,6 +36,8 @@ procs = cf
 
 [cf]
 # cf-store application
+# a space-separated list of infotags to set as CF Properties
+#infotags = archive foo bar blah
```
simon-ess: Inconsistent formatting: on this line `infotags` has no space before it, but on the other lines that should be uncommented there is a space. 🔧

mdavidsaver: This convention is intended to distinguish the line with the config file key from the explanatory comment.

simon-ess: I thought that might be the case, but it doesn't match the lines below. Either way seems good to me.

server/demo.conf
```diff
@@ -68,31 +67,39 @@ idkey = 42
...
+#recceiverID = recc1
+
 # Comma-separated list of VARIABLE:PropertyName,
 # specifying which environment VARIABLEs to pass on to the channel finder
 # and defining the corresponding PropertyName
 #environment_vars=ENGINEER:Engineer,EPICS_BASE:EpicsVersion,PWD:WorkingDirectory
+
 #Specify the timezone in the time output
-timezone = True
+#timezone = True
```
simon-ess, line 103: Why is this being removed? Also, this is also inconsistent formatting with the lines above 😄

server/demo.conf
```diff
@@ -26,13 +26,12 @@
...
 # Default plugins
 #  show - Prints information to daemon log
-#  db   - Stores in sqlite3 database
 #
 # eg. print and store in database.  Database config from section "[lite]"
 # Show config from "[show]" although show has no options at present.
 ##procs = show, db:lite
 
-procs = show
+procs = show, db:lite
```
simon-ess: Why is `db:lite` added, and where is it configured?

server/recceiver_full.conf
```diff
@@ -3,64 +3,90 @@
...
+#
+# Plugin names may be followed by an instance name (eg. "db:arbitrary")
+# which allows for more than one instance of a plugin with different
+# configuration
+#
+# Default plugins
+#  show - Prints information to daemon log
+#
+# eg. print and store in database.  Database config from section "[lite]"
+# Show config from "[show]" although show has no options at present.
```
simon-ess, line 28: `[lite]` section?

server/demo.conf
jacomago: Was keeping demo as the defaults only config file. So nothing is set, wheras recceiver_full.conf everything is set.

server/demo.conf
jacomago: Configured in the [lite] section.

server/recceiver_full.conf
jacomago: Added

server/demo.conf
jacomago: Defined lower down in the file in the [lite] section

server/recceiver_full.conf
```diff
@@ -0,0 +1,107 @@
...
+# Specify an optional id for the recceiver to be used with cleanOnStart and cleanOnStop
+# default value is the hostname of the machine the python interpreter is started on
+#recceiverID = recc1
+
+# Comma-separated list of VARIABLE:PropertyName,
+# specifying which environment VARIABLEs to pass on to the channel finder
+# and defining the corresponding PropertyName
+environment_vars=ENGINEER:Engineer,EPICS_BASE:EpicsVersion,PWD:WorkingDirectory
+
+#Specify the timezone in the time output
```
anderslindho: Formatting is a bit inconsistent throughout this file

anderslindho (approved):

PR #102, opened 2025-02-05
End to end testing for RecCeiver.

jacomago:
Add a smoke test for the full recsync+channelfinder pipeline.

Constructed via docker containers the test spins up a docker compose with

```
         cf
  |              | 
 recc1   recc2
|     |       |      |
ioc ioc ioc   ioc
```

Then makes sure the count of channels is the sum of all the channels in all the iocs


.gitignore
```diff
@@ -1,16 +1,11 @@
-O.*
```
anderslindho, line 1: The new gitignore setup feels a bit more complicated. If you wish to keep it, all common ignores should be in this root one.

.gitignore
```diff
@@ -1,16 +1,11 @@
...
-*.pyo
-*.log
-*.pid
-*.db
-client/configure/*.local
+# python
+.venv
 
-*~
-.*.swp
```
anderslindho: These should prob be kept under the IDE section. I would also add DS_Store.

.gitignore
```diff
@@ -1,16 +1,11 @@
...
+application.log
 
-*.pyc
-*.pyo
-*.log
-*.pid
-*.db
-client/configure/*.local
+# python
+.venv
```
anderslindho, line 4: IMO `venv` (without a leading period) is one of the most common names

client/.gitignore
```diff
@@ -0,0 +1,19 @@
...
+dropin.cache
+
+*.pyc
+*.pyo
+*.log
+*.pid
+configure/*.local
+
+*~
+.*.swp
```
anderslindho: Should only be in the root gitignore

client/Dockerfile
```diff
@@ -0,0 +1,57 @@
+# Download Epics
```
anderslindho, line 1: Feels a bit heavy that this generic container downloads EPICS base, and a specific version at that

client/demoApp/Db/Makefile
```diff
@@ -11,7 +11,7 @@ include $(TOP)/configure/CONFIG
 # Create and install (or just install) into <top>/db
 # databases, templates, substitutions like this
 DB += somerecords.db
-
+DB += archive.db
```
anderslindho: Unnecessary (remove to get a more concise diff)

client/demoApp/Db/somerecords.db
```diff
@@ -1,11 +1,12 @@
 
-record(longin, "$(P)li") {
+record(longin, "$(P):li") {
   alias("$(P)lix1")
   alias("$(P)lix2")
   info("test", "testing")
 }
 
-record(longout, "$(P)lo") {
+record(longout, "$(P):lo") {
```
anderslindho: Unnecessary

client/demoApp/src/demoMain.cpp
```diff
@@ -19,5 +19,5 @@ int main(int argc,char *argv[])
     }
     iocsh(NULL);
     epicsExit(0);
-    return(0);
+    return 0;
```
anderslindho: Unnecessary

server/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM python:3.11 AS build
+
+WORKDIR /home/recceiver
+
+COPY . ./server
+
+RUN python -m venv venv
```
anderslindho, line 7: Don't really need a virtual environment in the container

server/docker-compose.yml
```diff
@@ -0,0 +1,156 @@
+services:
```
anderslindho: I think this file should be made clear that it is build for system/integration tests, either by moving or renaming

server/setup.py
```diff
@@ -1,5 +0,0 @@
-#!/usr/bin/env python
-
-from setuptools import setup
-
-setup()
```
anderslindho: Hmmm, I believe this drops support for older setuptools/python versions

server/tests/docker.py
```diff
@@ -0,0 +1,43 @@
+import logging
+import pytest
+from testcontainers.compose import DockerCompose
+
+from docker import DockerClient
+import pathlib
```
anderslindho: Would be nice to have isort activated in linting

server/tests/docker.py
```diff
@@ -0,0 +1,43 @@
...
+
+LOG: logging.Logger = logging.getLogger(__name__)
+
+
+def fullSetupDockerCompose() -> DockerCompose:
```
anderslindho: Why the odd naming style for this function?

server/tests/docker.py
```diff
@@ -0,0 +1,43 @@
...
+    return DockerCompose(
+        str(current_path.parent.resolve()),
+        compose_file_name=str(
+            current_path.parent.joinpath("docker-compose.yml").resolve()
+        ),
+        build=True,
+    )
+
+
+def log_container_logs(compose: DockerCompose) -> None:
```
anderslindho: `fetch_containers_and_log_logs`? 🔧

server/tests/test_e2e.py
```diff
@@ -0,0 +1,109 @@
...
+LOG: logging.Logger = logging.getLogger(__name__)
+
+MAX_WAIT_SECONDS = 180
+EXPECTED_DEFAULT_CHANNELS = 32
+
+
+def check_channel_count(cf_client, name="*", expected_count=EXPECTED_DEFAULT_CHANNELS):
+    channels = cf_client.find(name="*")
+    LOG.info("Found %s channels", len(channels))
+    return len(channels) == expected_count
```
anderslindho: This seems odd, that you set an expected default channel count which you know from somewhere else (please reference) in a global variable, and then you use that count as a default in an arg here. Does this really need to be overridable?

server/tests/test_e2e.py
```diff
@@ -0,0 +1,109 @@
...
+)
+
+LOG: logging.Logger = logging.getLogger(__name__)
+
+MAX_WAIT_SECONDS = 180
+EXPECTED_DEFAULT_CHANNELS = 32
+
+
+def check_channel_count(cf_client, name="*", expected_count=EXPECTED_DEFAULT_CHANNELS):
+    channels = cf_client.find(name="*")
```
anderslindho: You set `name` in the function args and then throw it away only to set the same query here

server/tests/test_e2e.py
```diff
@@ -0,0 +1,109 @@
...
+    LOG.info("Found %s channels", len(channels))
```
anderslindho: This feels like a development debug print that should be removed

server/tests/test_e2e.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+def wait_for_sync(cf_client, check):
+    seconds_to_wait = 1
+    total_seconds_waited = 0
+    while total_seconds_waited < MAX_WAIT_SECONDS:
+        try:
+            if check(cf_client):
+                break
+        except Exception as e:
```
anderslindho: Better to catch the known/expected exceptions 🔧 

You can keep this as a fallback if you also want to catch any unexpected ones, or cancel systemexit or keyboardinterrupt, etc.

server/tests/test_e2e.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+def check_channel_count(cf_client, name="*", expected_count=EXPECTED_DEFAULT_CHANNELS):
+    channels = cf_client.find(name="*")
+    LOG.info("Found %s channels", len(channels))
+    return len(channels) == expected_count
+
+
+def wait_for_sync(cf_client, check):
+    seconds_to_wait = 1
```
anderslindho: ?

server/tests/test_e2e.py
```diff
@@ -0,0 +1,109 @@
...
+@pytest.fixture(scope="class")
+def cf_client(setup_compose):  # noqa: F811
+    return create_client_and_wait(setup_compose)
+
+
+class TestE2E:
+    def test_smoke(self, cf_client) -> None:
+        """
+        Test that the setup in the docker compose creates channels in channelfinder
+        """
```
anderslindho: IMO test methods/functions should not have docstrings, as they are not part of any API

They can become a slight almost anti-pattern like here, where the test name is unclear because the test instead is described in the docstring (which is not displayed in e.g. test summaries). It would instead be better to have method names like `test_that_setup_in_compose_creates_channels_in_cf`.

server/Dockerfile
jacomago: Its an nice easy non OS specific way of getting twisted.

client/Dockerfile
jacomago: Can work on a simpler version maybe using epics-containers in a future version.

server/setup.py
jacomago: Do you know which?

server/tests/test_e2e.py
```diff
@@ -0,0 +1,93 @@
...
+                    LOG.debug(log.decode("utf-8"))
+            self.compose.stop()
+
+    def test_smoke(self) -> None:
+        """
+        Test that the setup in the docker compose creates channels in channelfinder
+        """
+        LOG.info("Waiting for channels to sync")
+        cf_host, cf_port = self.compose.get_service_host_and_port("cf")
+        cf_url = f"http://{cf_host if cf_host else 'localhost'}:{cf_port}/ChannelFinder"
```
simon-ess: Can `cf_host` ever be `None` (or something similar) here?

server/tests/test_e2e.py
jacomago: Renamed

server/tests/docker.py
jacomago: renamed

.gitignore
jacomago: Cleaned up a bit.

server/tests/test_e2e.py
```diff
@@ -0,0 +1,93 @@
...
+                conts = {
+                    container.ID: container
+                    for container in self.compose.get_containers()
+                }
+                for cont_id, cont in conts.items():
+                    log = docker_client.containers.get(cont_id).logs()
+                    LOG.debug("Info for container %s", cont)
+                    LOG.debug("Logs for container %s", cont.Name)
+                    LOG.debug(log.decode("utf-8"))
+            self.compose.stop()
```
simon-ess: Is there some reason to use this instead of fixtures?

jacomago: I do use a fixture in the next commit 😁

server/tests/test_e2e.py
jacomago: I guess it would just crash... But certainly the test would fail.

server/tests/test_e2e.py
simon-ess: 🤦 I guess that's what I get for forgetting that I was looking at a single commit...

server/tests/test_e2e.py
simon-ess: I guess my point is more... why have the `if.. else` clause at all? `cf_host` is defined on the line above, so unless you know that it can return a false-y object, why even check conditionally?

server/setup.py
anderslindho: Not sure but I would guess that there's a breaking point around py3.6, so maybe setuptools ~42 (was that maybe when experimental support for pyproject.toml for all config was added?).

I have a faint recollection that you also needed setup.py for editable installs (`pip install -e .`)

server/tests/test_e2e.py
jacomago: Ah, I see now. It is None locally (for me I think), but in an environment where the docker engine is running somewhere else it can be something else.

server/Dockerfile
anderslindho: Hmm, why? Can't you just do a user - or even a root - install?

server/.gitignore
```diff
@@ -0,0 +1,83 @@
+# Byte-compiled / optimized / DLL files
```
anderslindho, line 1: You are already covering a lot of this and of the distro/packaging entries in the root level gitignore through pyo, pyc ignore

server/.gitignore
```diff
@@ -0,0 +1,83 @@
...
+*.mo
+*.pot
+
+
+# PEP 582; used by e.g. github.com/David-OConnor/pyflow and github.com/pdm-project/pdm
+__pypackages__/
+
+# Environments
+.env
+.venv
```
anderslindho: In the root gitiginore

server/.gitignore
```diff
@@ -0,0 +1,83 @@
...
+.mypy_cache/
+.dmypy.json
+dmypy.json
+
+# PyCharm
+#  JetBrains specific template is maintained in a separate JetBrains.gitignore that can
+#  be found at https://github.com/github/gitignore/blob/main/Global/JetBrains.gitignore
+#  and can be added to the global gitignore or merged into this file.  For a more nuclear
+#  option (not recommended) you can uncomment the following to ignore the entire idea folder.
+.idea/
```
anderslindho: In the root gitignore

server/Dockerfile
simon-ess: Some folks advocate for this as a general thing to do: https://hynek.me/articles/docker-virtualenv/ 🤷

anderslindho (approved):
Largely fine now. Still some concerns about removing setup.py as part of this PR though as you're dropping unknown backwards compatability.

server/.gitignore
jacomago: I think you're confused with the client one.

PR #104, opened 2025-03-10
Rename shorter variables to longer more descriptive ones

jacomago:


server/recceiver/scripts/add_extra_properties.py
```diff
@@ -1,7 +1,7 @@
 from channelfinder import ChannelFinderClient
 
 """
-Simple script for adding active channels to Channel Finder Service for testing cf-store clean
```
anderslindho: channels actually seems appropriate here

server/recceiver/scripts/print_cf_data.py
```diff
@@ -8,29 +8,29 @@
...
-        ch.pop("owner", None)
-        ch.pop("tags", None)
-        for prop in ch["properties"]:
-            if prop["name"] == "hostName":
-                ch["hostName"] = prop["value"]
-            if prop["name"] == "iocName":
-                ch["iocName"] = prop["value"]
-        ch.pop("properties", None)
-    return channels
+    for cf_record in records:
```
anderslindho: same here

server/recceiver/scripts/readme
```diff
@@ -17,7 +17,7 @@ Clean and poll on recsync start:
     - polling completes
 4. CF will clean
 5. Check result
-    - all channels inactive
+    - all records inactive
```
anderslindho: and so on 😬 

maybe drop commit(s) converting channels to records and similar

server/recceiver/cfstore.py
```diff
@@ -213,14 +213,14 @@ def _commitWithThread(self, transaction):
         """The unique identifier for a particular IOC"""
         iocid = host + ":" + str(port)
 
-        pvInfo = {}
+        recordInfo = {}
```
anderslindho, line 216: Maybe fix the naming style in the same go?

anderslindho (approved):
Some seem a little bit overkill to me, like `cf_properties` and `cf_channels` (should it then also be `new_cf_channels`, etc?), but overall fine. Will approve but would be nice if you fixed the camelCase var I dropped a comment on

Issue #92, opened 2024-06-20
Info tags persist after recceiver restart with new configuration


jacomago:
Should be reproducible in the following ways:

    Create recceiver with infotag (X) set to be updated in channelfinder.
    Add IOC with PV with infotag X set to A
    Channelfinder has PV with X=A
    Reboot recceiver with  infotag X removed
    Channelfinder has PV with X=A still even after IOC update

    Create recceiver with infotag (X) set to be updated in channelfinder.
    Add IOC with PV with infotag X set to A
    Channelfinder has PV with X=A
    Reboot IOC with PV with infotag X removed
    Channelfinder has PV with X=A still

Possible solution is to set in recceiver to remove tags and properties from Channelfinder PVs if recceiver is the owner. (probably behind a configuration flag)

tynanford:
> Possible solution is to set in recceiver to remove tags and properties from Channelfinder PVs if recceiver is the owner. (probably behind a configuration flag)

+1 this seems like a nice feature

anderslindho:
Wasn't this fixed @jacomago?

PR #105, opened 2025-03-14
Fix infotag loss bug

jacomago:
Adds a test for and fixes issue #92


server/recceiver/cfstore.py
```diff
@@ -401,6 +395,46 @@ def dict_to_file(dict, iocs, conf):
...
+        "properties": properties,
+    }
+
+
+def create_property(owner: str, name: str, value: str):
+    return {
+        "name": name,
+        "owner": owner,
+        "value": value,
+    }
```
simon-ess, line 412: Extremely naïve question, but this looks a lot like you just want a dataclass, doesn't it? Or is that a "python is too old issue" or something else?

server/recceiver/cfstore.py
```diff
@@ -710,14 +710,14 @@ def create_default_properties(owner, iocTime, recceiverid, channels_dict, iocs,
...
-def __merge_property_lists(newProperties, oldProperties):
+def __merge_property_lists(newProperties: list[dict[str, str]], channel: dict[str, list[dict[str, str]]]):
     """
     Merges two lists of properties ensuring that there are no 2 properties with
     the same name In case of overlap between the new and old property lists the
     new property list wins out
     """
     newPropNames = [p["name"] for p in newProperties]
-    for oldProperty in oldProperties:
+    for oldProperty in channel["properties"]:
```
simon-ess, line 751: Is there any chance that `properties` does not exist as a key?

jacomago:
Current problem with this version is that owner is specified twice in cfstore, once from config, another from the ioc itself. Kind of makes sense, but means my check on the properties doesn't work in the merge_property_list. Need a cleverer check for the owner.

jacomago:
Realised should have all of the property names that receiver deals with anyway. The required property list. Just need to pass that as an extra argument

server/recceiver/recast.py
```diff
@@ -241,6 +241,27 @@ def __str__(self):
...
+            records_to_delete={})
+            """.format(
+            self.source_address,
+            self.initial,
+            self.connected,
+            self.records_to_add,
+            self.client_infos,
+            self.record_infos_to_add,
+            self.aliases,
+            self.records_to_delete,
```
anderslindho: Maybe change this to use an f-string (for readability), and then use `self.__class__.__name__` for the class name (for maintainability)? It would be nice if we also could do `{var=}`, but unfortunately that seems to print also `self` in this case.

server/test-remove-compose.yml
```diff
@@ -0,0 +1,88 @@
+name: recceiver-test-remove-compose
```
anderslindho: Maybe you could use include (https://docs.docker.com/compose/how-tos/multiple-compose-files/include/) for parts of the compose setup

server/tests/client_checks.py
```diff
@@ -0,0 +1,74 @@
...
+INACTIVE_PROPERTY = {"name": "pvStatus", "owner": "admin", "value": "Inactive", "channels": []}
+MAX_WAIT_SECONDS = 180
+TIME_PERIOD_INCREMENT = 2
+
+
+def channel_match(channel0, channel1, properties_to_match: list[str]):
+    if channel0["name"] != channel1["name"]:
+        assert False, f"Channel names do not match: {channel0['name']} != {channel1['name']}"
+    if channel0["owner"] != channel1["owner"]:
+        assert False, f"Channel owners do not match: {channel0['owner']} != {channel1['owner']}"
```
anderslindho: You can just do `assert condition, message`

server/tests/client_checks.py
```diff
@@ -0,0 +1,74 @@
...
+    for prop in channel0["properties"]:
+        if prop["name"] in properties_to_match:
+            if prop not in channel1["properties"]:
+                assert False, f"Property {prop} not found in channel {channel1['name']}"
```
anderslindho: If you skip the log message you can simplify this to an `all` or `any` check

server/tests/client_checks.py
```diff
@@ -0,0 +1,74 @@
...
+    assert True
+
+
+def channels_match(channels_begin, channels_end, properties_to_match: list[str]):
+    for channel_index, channel in enumerate(channels_begin):
+        channel_match(channel, channels_end[channel_index], properties_to_match)
+
+
+def check_channel_count(cf_client: ChannelFinderClient, name="*", expected_channels=24):
```
anderslindho: Should be `expected_channel_count`

server/tests/client_checks.py
```diff
@@ -0,0 +1,74 @@
...
+    LOG.debug("Found %s channels", len(channels))
+    return len(channels) == expected_channels
+
+
+def check_channel_property(cf_client: ChannelFinderClient, name="*", prop=ACTIVE_PROPERTY):
+    channels = cf_client.find(name=name)
+    for channel in channels:
+        if prop not in channel["properties"]:
+            return False
+    return True
```
anderslindho:
```python
return all(prop in channel["properties"] for channel in channels)
```

server/tests/test_remove_property.py
```diff
@@ -0,0 +1,95 @@
...
+    LOG.info("Stopping docker compose")
+    if LOG.level <= logging.DEBUG:
+        fetch_containers_and_log_logs(compose)
+    compose.stop()
+
+
+def docker_exec_new_command(container: Container, command: str):
+    def stream_logs(exec_result, cmd: str):
+        if LOG.level <= logging.DEBUG:
+            LOG.debug("Logs from ioc1-1 with command %s", cmd)
```
anderslindho: Should this really have a hardcoded IOC-name?

server/tests/test_remove_property.py
```diff
@@ -0,0 +1,95 @@
...
+        target=stream_logs,
+        args=(
+            exec_result,
+            command,
+        ),
+    )
+    log_thread.start()
+
+
+class TestRemoveProperty:
```
anderslindho, line 54: Why a class here?

server/tests/test_remove_property.py
```diff
@@ -0,0 +1,95 @@
...
+
+        # Check ioc1-1 has ai:archive with info tag "archive"
+        LOG.debug('Checking ioc1-1 has ai:archive with info tag "archive"')
+        archive_channel_name = "IOC1-1:ai:archive"
+        archive_channel = cf_client.find(name=archive_channel_name)
+
+        def get_len_archive_properties(archive_channel):
+            return len([prop for prop in archive_channel[0]["properties"] if prop["name"] == "archive"])
+
+        assert get_len_archive_properties(archive_channel) == 1
```
anderslindho, line 75: I guess you are using a named function to document, but I am not sure it's needed here. Don't care hugely though.

server/tests/test_restart.py
```diff
@@ -0,0 +1,47 @@
...
+from .docker import restart_container, setup_compose  # noqa: F401
+
+PROPERTIES_TO_MATCH = ["pvStatus", "recordType", "recordDesc", "alias", "hostName", "iocName", "recceiverID"]
+
+LOG: logging.Logger = logging.getLogger(__name__)
+
+EXPECTED_DEFAULT_CHANNELS = 32
+
+
+@pytest.fixture(scope="class")
```
anderslindho, line 17: You could also do module scope and skip the class, since it's only two tests in this file

anderslindho (commented):
Great find and fix!

Maybe separete the fix from the test that catches it in the git history. 🔧 

It's a bit annoying how hardcoded much of the testing is here...

server/recceiver/cfstore.py
```diff
@@ -124,6 +124,7 @@ def _startServiceWithLock(self):
                     self.client.set(property={"name": cf_property, "owner": owner})
 
                 self.record_property_names_list = set(record_property_names_list)
+                self.managed_properties = required_properties.union(set(record_property_names_list))
```
simon-ess: Do you need the `set` here? I think you can union any iterable... 🔧

server/recceiver/cfstore.py
```diff
@@ -632,45 +656,53 @@ def __updateCF__(
...
+            _log.debug(
+                "Update existing channel with different IOC 0: {s} newProps {n}".format(s=existingChannel, n=newProps)
+            )
+            existingChannel["properties"] = __merge_property_lists(
+                newProps,
+                existingChannel,
+                processor.managed_properties,
+            )
+            _log.debug(
+                "Update existing channel with different IOC 1: {s} newProps {n}".format(s=existingChannel, n=newProps)
```
simon-ess: These look like weird debug messages to me with the 0/1 in them. I guess you're doing a before/after on `existingChannel`, but maybe a litteral before/after would be clearer, e.g.
```py
log.debug("Updating existing channel {existingChannel} with {newProps}")
# update
log.debug("After update: {existingChannel}")
```

server/tests/client_checks.py
simon-ess: By jove, he's right!
```
>>> assert False, "foo"
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AssertionError: foo
>>> assert True, "bar"
>>>
```

server/tests/test_e2e.py
```diff
@@ -1,73 +1,32 @@
...
-
-
-def create_client_from_compose(compose):
-    cf_host, cf_port = compose.get_service_host_and_port("cf")
-    cf_url = f"http://{cf_host if cf_host else 'localhost'}:{cf_port}/ChannelFinder"
-    # wait for channels to sync
-    LOG.info("CF URL: %s", cf_url)
-    cf_client = ChannelFinderClient(BaseURL=cf_url)
-    return cf_client
+EXPECTED_DEFAULT_CHANNELS = 32
```
simon-ess: Pedantic, but mabye this should be `EXPECTED_DEFAULT_CHANNEL_COUNT`; channels makes it sounds like a list of channels. 🔧

server/recceiver/cfstore.py
anderslindho: You could also use the union operator `|` 🔧

server/tests/client_checks.py
```diff
@@ -0,0 +1,72 @@
...
+INACTIVE_PROPERTY = {"name": "pvStatus", "owner": "admin", "value": "Inactive", "channels": []}
+MAX_WAIT_SECONDS = 180
+TIME_PERIOD_INCREMENT = 2
+
+
+def channel_match(channel0, channel1, properties_to_match: list[str]):
+    assert channel0["name"] == channel1["name"], f"Channel names do not match: {channel0['name']} != {channel1['name']}"
+    assert channel0["owner"] != channel1["owner"], (
+        f"Channel owners do not match: {channel0['owner']} != {channel1['owner']}"
+    )
```
anderslindho: This check is the opposite of what the message says

Also, do you need to write out the condition and the values - aren't these in the stack trace anyway?

server/tests/client_checks.py
```diff
@@ -0,0 +1,72 @@
...
+
+    for prop in channel0["properties"]:
+        assert not (prop["name"] in properties_to_match and prop not in channel1["properties"]), (
+            f"Property {prop} not found in channel {channel1['name']}"
+        )
+    assert True
```
anderslindho: This line does nothing

server/recceiver/cfstore.py
jacomago: removed

server/tests/test_remove_property.py
jacomago: Eh, why not

server/recceiver/cfstore.py
jacomago: More I thought adding it would be another refactor.

server/recceiver/cfstore.py
jacomago: Not here... I hope. I think raising exception is pretty suitable response if it happens. Which is the current behaviour.

server/recceiver/cfstore.py
jacomago: fixed

anderslindho (approved):

PR #106, opened 2025-03-21
Add upload of docker image

jacomago:
Add upload of the recciever image so it can't be pulled in other contexts.


anderslindho (approved):
LGTM but something to explain the reason for the MR (in the description or so) would have been nice

PR #107, opened 2025-04-02
Fix docker upload

jacomago:


simon-ess:
LGTM, I guess? However the new pipeline isn't running as the current changes aren't in `server`...

EDIT: Wait, the README is there...

jacomago:
> LGTM, I guess? However the new pipeline isn't running as the current changes aren't in `server`...
> 
> EDIT: Wait, the README is there...

I think its because its a fork... 

And its not on pull requests.

I did it that way because it makes more sense to push the docker image to jacomago/recsync/recceiver than channelfinder/recsync/recceiver-someprname

shroffk (approved):

PR #109, opened 2025-06-17
Re-enable checks for cf conf keys to be set to on

tynanford:
Version 1.7 of recsync changed how the CF configuration settings work. Now if you set `recordDesc`, `recordType`, `iocConnectionInfo`, or `alias` in the cf.conf file then they automatically are enabled. The code doesn't check if the setting is "on" or "off". In addition `iocConnectionInfo` was supposed to be on by default but with this change now you have to define it in the configuration file.

This adds back the checks for these properties and forces the user to set `recordDesc`, `alias`, and `recordType` explicitly to "on" like before.

### DIFF
--- a/server/demo.conf
+++ b/server/demo.conf
@@ -69,6 +69,7 @@
 #infotags = archive foo bar blah
 
 # Uncomment line below to turn off the feature to add CA/PVA port info for name server to channelfinder
+# By default this is enabled
 #iocConnectionInfo = off
 
 # Uncomment line below to turn on the feature to add alias records to channelfinder
--- a/server/recceiver/cfstore.py
+++ b/server/recceiver/cfstore.py
@@ -88,9 +88,9 @@ class CFProcessor(service.Service):
                     RECCEIVERID_KEY,
                 }
 
-                if self.conf.get("alias"):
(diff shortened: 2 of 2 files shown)


jacomago:
Why explicitly check if the field is "on" rather than just set like normal boolean flags?

Also, the reason I removed it because it was inconsistent throughout, whether the flag was set vs "on"

tynanford:
> Why explicitly check if the field is "on" rather than just set like normal boolean flags?

Because this is how recceiver configuration has been working since 2018 https://github.com/ChannelFinder/recsync/pull/33

`iocConnectionInfo` was originally set to be enabled by default. This is changed now in 1.7 and you have to uncomment the setting in the conf file to get it to work. And confusingly uncommenting this line _enables_ iocConnectionInfo

```
# Uncomment line below to turn off the feature to add CA/PVA port info for name server to channelfinder
#iocConnectionInfo = off
```

Right now you can only disable the setting by either commenting it out or setting it to a blank string. It looks like using `self.conf.getboolean` works well though and is backwards compatible so I will close this PR in favor of https://github.com/ChannelFinder/recsync/pull/110

PR #110, opened 2025-06-18
Update alias, recordType, recordDesc, and iocConnectionInfo settings to be booleans

tynanford:
Alternate version of https://github.com/ChannelFinder/recsync/pull/109

There is a nice side-effect with the python ConfigParser.getboolean function where it is case-insensitive and recognizes Boolean values from `yes`/`no`, `on`/`off`, `true`/`false` and `1`/`0` https://docs.python.org/3/library/configparser.html#supported-datatypes

So older setting files like this would still work as before:

```
alias = off
recordType = on
```

### DIFF
--- a/server/demo.conf
+++ b/server/demo.conf
@@ -69,16 +69,16 @@
 #infotags = archive foo bar blah
 
 # Uncomment line below to turn off the feature to add CA/PVA port info for name server to channelfinder
-#iocConnectionInfo = off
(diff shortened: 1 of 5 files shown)


jacomago:
Looks way better.

jacomago (approved):

simon-ess:
I agree with @jacomago

simon-ess (approved):

PR #112, opened 2025-07-25
Reccaster exclusion pattern for record filtering

madelinespark:


client/castApp/src/castinit.c
```diff
@@ -200,6 +200,127 @@ static void addReccasterEnvVarsCallFunc(const iocshArgBuf *args)
...
+    argv++; argc--; /* skip function arg */
+    if(argc < 1) {
+        errlogSevPrintf(errlogMinor, "At least one argument expected for addReccasterExclusionPattern\n");
+        return;
+    }
+
+    // error if called after iocInit()
+    if(self->current != casterStateInit) {
+        errlogSevPrintf(errlogMinor, "addReccasterExcludePattern called after iocInit() when reccaster might already be connected. Not supported\n");
+        epicsMutexUnlock(self->lock);
```
simon-ess, line 223: The mutex is not locked yet.

client/castApp/src/castinit.c
```diff
@@ -200,6 +200,127 @@ static void addReccasterEnvVarsCallFunc(const iocshArgBuf *args)
...
+            num_valid_args--;
+        }
+        else {
+            for (j = i + 1; j < argc; j++) {
+                if (argv[j] && (strcmp(argv[i], argv[j]) == 0)) {
+                    errlogSevPrintf(errlogMinor, "Duplicate pattern %s\n", argv[i]);
+                    argv[i] = NULL;
+                    num_valid_args--;
+                }
+            }
```
simon-ess: This is more of a discussion than anything, but this looks so odd to me. I would expect that you would check _later_ arguments against _previous_ ones, but this checks the _current_ one if _any future ones_ match it. I guess it's the same, but it breaks my brain a bit.

(If you set it to `NULL` though, you should be able to break out of it...)

client/castApp/src/castinit.c
```diff
@@ -200,6 +200,127 @@ static void addReccasterEnvVarsCallFunc(const iocshArgBuf *args)
...
+    /* check for duplicates between argv and existing */
+    for (i = 0; i < argc; i++) {
+        for (j = 0; j < self->num_exclude_patterns; j++) {
+            if (argv[i] && strcmp(argv[i], self->exclude_patterns[j]) == 0) {
+                errlogSevPrintf(errlogMinor, "Pattern %s is already in exclude_patterns list\n", argv[i]);
+                argv[i] = NULL;
+                num_valid_args--;
+            }
+        }
+    }
```
simon-ess: Why not do this at the same time as the above?

client/castApp/src/castinit.c
```diff
@@ -200,6 +200,127 @@ static void addReccasterEnvVarsCallFunc(const iocshArgBuf *args)
...
+                if (argv[count] != NULL) {
+                    if ((new_exclude[i] = strdup(argv[count])) == NULL) {
+                        errlogSevPrintf(errlogMinor, "strdup error for copying %s to new_exclude[%zu] from addReccasterExcludePattern\n", argv[i], i);
+                        ret = 1;
+                        break;
+                    }
+                    i++;
+                }
+            }
+        }
```
simon-ess: I think that a lot of this code could be simplified by using EPICS linked list structures instead of manually managing arrays and copying them over; for example, the initial function call could just be:
```
for (int i = 0; i < argc; i++) {
  if (argv[i] && argv[i][0]) {
    if (argv[i] not in linked_list) {
      linked_list.append(argv[i]);
    }
  }
}
```
(give or take)

client/castApp/src/castinit.c
```diff
@@ -200,6 +200,127 @@ static void addReccasterEnvVarsCallFunc(const iocshArgBuf *args)
...
+    epicsMutexUnlock(self->lock);
+}
+
+static const iocshArg addReccasterExcludePatternArg0 = { "excludePattern", iocshArgArgv };
+static const iocshArg * const addReccasterExcludePatternArgs[] = { &addReccasterExcludePatternArg0 };
+static const iocshFuncDef addReccasterExcludePatternFuncDef = {
+    "addReccasterExcludePattern",
+    1,
+    addReccasterExcludePatternArgs
+};
```
simon-ess, line 262: Another thought: maybe this shouldn't be a varargs-type function, but take a single argument. In my opinion, it is a lot cleaner and easier to follow if someone has
```
addReccasterExcludePattern PAT1
addReccasterExcludePattern PAT2
addReccasterExcludePattern PAT3
```
instead of
```
addReccasterExcludePattern PAT1 PAT2 PAT3
```

client/castApp/src/dbcb.c
```diff
@@ -57,6 +61,9 @@ static int pushEnv(caster_t *caster)
         }
         free(buf);
     }
+    // if(!getenv("RECCASTER_EXCLUDE")) {
+    //     epicsEnvSet("RECCASTER_EXCLUDE", "");
+    // }
```
simon-ess: Commented code.

simon-ess (changes requested):

client/castApp/src/castinit.c
madelinespark: fixed!

client/castApp/src/dbcb.c
madelinespark: fixed!

client/castApp/src/castinit.c
madelinespark: fixed!

client/castApp/src/castinit.c
madelinespark: changed this

client/castApp/src/castinit.c
tynanford: the varargs gives you the option to do either and we don't have a strong preference either way. This currently matches how `addReccasterEnvVars` works which was the reasoning for varargs in `addReccasterExcludePattern` . But no problem if we should change to a single arg per call

client/castApp/src/castinit.c
tynanford: Hi Simon, are you referring to `ELLLIST` in `ellLib.h` ?

simon-ess: That's the one, yes.

client/castApp/src/castinit.c
simon-ess: That's fair. I think I don't love varargs unless it makes obvious sense (like `echo` which... doesn't use varargs 😄 ) and would rather one per line. But you are correct that this is how it works for the env vars, so I am happy to leave this up to your discretion.

tynanford: Sounds good. We will leave with varargs to match addReccasterEnvVars unless someone else has objections

client/castApp/src/castinit.c
```diff
@@ -200,6 +200,84 @@ static void addReccasterEnvVarsCallFunc(const iocshArgBuf *args)
...
+    if(self->current != casterStateInit) {
+        errlogSevPrintf(errlogMinor, "addReccasterExcludePattern called after iocInit() when reccaster might already be connected. Not supported\n");
+        epicsMutexUnlock(self->lock);
+        return;
+    }
+
+    for (i = 0; i < argc; i++) {
+        if (argv[i] == NULL) {
+            errlogSevPrintf(errlogMinor, "Arg is NULL for addReccasterExcludePattern\n");
+        }
```
simon-ess: Honest question here: Can this actually happen? I would assume that the `argv`s will never be null, but I am not certain...

simon-ess (approved):

client/castApp/src/castinit.c
madelinespark: removed the null check as it seems like argv will never be null

madelinespark:
squashed all commits

tynanford:
@madelinespark and I looked through and cleaned up things a bit in `addReccasterExcludePattern`
 - removed a few nested if/else blocks in favor of `continue`
 -  removed the `ret` variable and extra logging message that just said there was an error in `addReccasterExcludePattern`
 - changed strcmp to `argv[i][0] == '\0'`

we were wondering if `ellFree` is cleaning all the memory or not. Madeline will check on that

tynanford (approved):
Looks good to me and tested with base R3.15.8 and R7.0.9. ELLLIST is very handy!

tynanford:
Hi @simon-ess , sorry for another ping but there have been enough changes since your last approval that we would appreciate one final look through

simon-ess (approved):
LGTM!

simon-ess:
Note that you're failing a sonarqube check for duplicated code on the test function. It might be good to refactor that code, but you also could consider disabling the check on test code?

tynanford:
I'd be in favor of removing that check on the test file.

@mdavidsaver  comments on https://github.com/ChannelFinder/recsync/pull/117#discussion_r2347028501 would apply to this PR as well. The PRs go in tandem and I had Madeline create the addReccasterEnvVars refactor PR on Friday since that was the last day of her internship. It branches off this PR since there is shared code like `nodeFree` and `string_list_t`

~~My proposal would be to merge this PR as is if it is acceptable and then refactor the loop iteration code and memory allocations for both addReccasterEnvVars and addReccasterExcludePattern in PR #117 . That way the git merging is less troublesome. But let us know if this PR should be cleaned up first.~~

edit - actually maybe it makes more sense to get this PR fully completed and then worry about pr #117

simon-ess:
Yeah, I saw that that other PR was based off of this one and sort of assumed that it would be rebased after this one is merge.

tynanford:
The changes Michael mentioned about array allocation and looping through the linked lists are implemented. Also updated the variable naming as suggested. This is ready to merge from our end, can you take one last look @simon-ess  ?

simon-ess (approved):

PR #113, opened 2025-08-06
Add db dir to client gitignore and standardize a bit

tynanford:


simon-ess (approved):
Are there any other directories we should add at the same time? I see dbd, bin, lib, so I guess that's it?

tynanford:
It seems like most other EPICS modules have the include directory as well so I added that. Also went ahead and specified which things in the gitignore are directories. And deleted references to python stuff which should just be in `server/` gitignore file

simon-ess (approved):

PR #118, opened 2025-09-25
switch type hints to be py36 compatible

tynanford:
https://github.com/ChannelFinder/recsync/pull/105 added some type hint syntax which is only available on python 3.9+ . This updates it to be backward compatible to the target 3.6 version

Also added 3.6 and 3.7 to the CI with a docker container since there aren't any ubuntu runners that support python < 3.8  - https://github.com/actions/setup-python/issues/1048 . And added 3.13 to the build-server CI

~~I'm not sure why the CI for the container tests are failing and if it's related to this or not.~~ Re-ran the failed tests and it worked

### DIFF
--- a/.github/workflows/server.yml
+++ b/.github/workflows/server.yml
@@ -14,15 +14,27 @@ on:
 jobs:
   build-server:
     runs-on: ubuntu-latest
+    container: ${{ matrix.container }}
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        include:
+          - python-version: "3.6"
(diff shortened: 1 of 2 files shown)


simon-ess (approved):
I would like to say we should drop py3.6, but given that we also run a Centos7 facility here, I think that would be shooting ourselves in the food :smile:

tynanford:
sounds like we need to add python 2.7 to the CI then! :joy:

simon-ess:
😱

jacomago:
Actually we use a python 3.13 container... So at ESS we can remove py36

PR #119, opened 2025-10-06
Chunk the calls to ChannelFinder

jacomago:
This solves an issue we had here with a 170 000 PV IOC where the ChannelFinder errored out.

### DIFF
--- a/server/demo.conf
+++ b/server/demo.conf
@@ -80,7 +80,8 @@
 # Uncomment line below to turn on the feature to add description field to channelfinder
 #recordDesc = True
 
-# The size limit for finding channels (ie the value of the '~size' query parameter)
+# The size limit for finding channels and submitting channels
+# (ie the value of the '~size' query parameter)
 # If not specified then the fallback is the server default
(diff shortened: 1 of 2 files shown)


simon-ess (changes requested):
Seems reasonable, although the configuration parameter should be documented somewhere.

Also, the tests are failing for some reason.

shroffk (approved):

anderslindho (approved):

tynanford (approved):
some of the recceiver tests seem flaky. Re-ran the test and the third time was the charm

mdavidsaver:
Does this change update any minimum version requirements?  CF client library?  CF service?

jacomago:
> Does this change update any minimum version requirements?  CF client library?  CF service?
Nope. No changes there.

shroffk:
FYI:
https://github.com/ChannelFinder/ChannelFinderService/pull/187

Adding support on the service side to help address this problem

simon-ess (approved):

PR #117, opened 2025-09-12
changed addReccasterEnvVars to linked list implementation

madelinespark:
NOTE: https://github.com/ChannelFinder/recsync/pull/112 needs to be merged first

PR #120, opened 2025-10-06
Convert addReccasterEnvVars to linked list and create helper function for addReccaster* logic

tynanford:
This PR is in place of https://github.com/ChannelFinder/recsync/pull/117 so the git merging/commit history isn't messed up. I branched off the latest in master, cherry-picked @madelinespark 's two commits, and worked off that.

- Updated the memory allocation and looping to match this suggestion https://github.com/ChannelFinder/recsync/pull/117#discussion_r2347028501
- Created function to handle both `addReccasterEnvVars` and `addReccasterExcludePattern` logic
- Added `IOCSHFUNCDEF_HAS_USAGE` message for `addReccasterExcludePattern`

Madeline's commits already converted `addReccasterEnvVars` to a linked list implementation and updated all the unit tests.

### DIFF
--- a/client/castApp/src/caster.c
+++ b/client/castApp/src/caster.c
@@ -114,6 +114,7 @@ void casterInit(caster_t *self)
     self->onmsg = &casterShowMsgDefault;
     self->current = casterStateInit;
     self->timeout = reccastTimeout;
+    ellInit(&self->extra_envs);
     ellInit(&self->exclude_patterns);
 
     if(shSocketPair(self->wakeup))
@@ -122,7 +123,6 @@ void casterInit(caster_t *self)
 
 void casterShutdown(caster_t *self)
 {
-    int i;
     epicsUInt32 junk = htonl(0xdeadbeef);
 
     epicsMutexMustLock(self->lock);
@@ -137,10 +137,7 @@ void casterShutdown(caster_t *self)
     epicsEventMustWait(self->shutdownEvent);
 
     epicsMutexMustLock(self->lock);
-    for (i = 0; i < self->num_extra_envs; i++) {
-        free(self->extra_envs[i]);
-    }
-    free(self->extra_envs);
+    ellFree2(&self->extra_envs, &nodeFree);
     epicsMutexUnlock(self->lock);
 
     epicsMutexMustLock(self->lock);
--- a/client/castApp/src/caster.c
+++ b/client/castApp/src/caster.c
@@ -7,6 +7,7 @@
 #include <epicsAssert.h>
 #include <epicsThread.h>
 #include <epicsStdio.h>
+#include <cantProceed.h>
 
 #define epicsExportSharedSymbols
 
@@ -114,8 +115,11 @@ void casterInit(caster_t *self)
     self->onmsg = &casterShowMsgDefault;
     self->current = casterStateInit;
     self->timeout = reccastTimeout;
-    ellInit(&self->extra_envs);
+    ellInit(&self->envs);
     ellInit(&self->exclude_patterns);
+
+    /* add default_envs to envs list which can be expanded by the user with addReccasterEnvVars iocsh function */
+    addToReccasterLinkedList(self, default_envs_count, default_envs, &self->envs, "casterInit", "Default environment variable");
 
     if(shSocketPair(self->wakeup))
         errlogPrintf("Error: casterInit failed to create shutdown socket: %d\n", SOCKERRNO);
@@ -137,7 +141,7 @@ void casterShutdown(caster_t *self)
     epicsEventMustWait(self->shutdownEvent);
 
     epicsMutexMustLock(self->lock);
-    ellFree(&self->extra_envs);
+    ellFree(&self->envs);
     ellFree(&self->exclude_patterns);
     epicsMutexUnlock(self->lock);
 
--- a/client/castApp/src/caster.c
+++ b/client/castApp/src/caster.c
@@ -7,7 +7,6 @@
 #include <epicsAssert.h>
 #include <epicsThread.h>
 #include <epicsStdio.h>
-#include <cantProceed.h>
 
 #define epicsExportSharedSymbols
 
--- a/client/castApp/src/caster.c
+++ b/client/castApp/src/caster.c
@@ -137,10 +137,7 @@ void casterShutdown(caster_t *self)
     epicsEventMustWait(self->shutdownEvent);
 
     epicsMutexMustLock(self->lock);
-    ellFree2(&self->extra_envs, &nodeFree);
-    epicsMutexUnlock(self->lock);
-
-    epicsMutexMustLock(self->lock);
+    ellFree(&self->extra_envs);
     ellFree(&self->exclude_patterns);
     epicsMutexUnlock(self->lock);
 
--- a/client/castApp/src/castinit.c
+++ b/client/castApp/src/castinit.c
@@ -76,20 +76,19 @@ static void casthook(initHookState state)
 }
 
 /* Helper function to add items from iocsh calls to internal linked lists
+ * self is the caster instance
+ * itemCount is the number of items in the items array
+ * items is the array of strings to add to the list
+ * reccastList is the linked list to add the items to
  * funcName is the name of the IOC shell function being called (for error messages)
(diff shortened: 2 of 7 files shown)


client/castApp/src/castinit.c
```diff
@@ -75,21 +75,21 @@ static void casthook(initHookState state)
...
-    Example call: addReccasterEnvVars("SECTOR") or addReccasterEnvVars("SECTOR", "BUILDING")
-    Appends the given env variables to the extra_envs list to be sent in addition to the default_envs array
-*/
-void addReccasterEnvVars(caster_t* self, int argc, char **argv)
+/* Helper function to add items from iocsh calls to internal linked lists
+ * funcName is the name of the IOC shell function being called (for error messages)
+ * itemDesc is string to describe what is being added (for error messages)
+ * defaultArray is an optional arg for a default list to also check for duplicates
+ */
+static void addToReccasterLinkedList(caster_t* self, int argc, char **argv, ELLLIST* list, const char* funcName, const char* itemDesc, const char** defaultArray)
```
simon-ess: I can see why this signature is used, although I don't love it.

One thought that might clean up a bunch of the code and improve this: at the moment for the default_env array we treat this as separate from the defined ones. Why not start the caster initialisation with the default array? That is, we could actually start with
```c
addToReccasterLinkedList(caster, count default_env_vars, addReccasterEnvVars", "Default Environment variable");
```
or something like that. This would allow you to get rid of the last argument as well as the `if(defaultArray)` block.

tynanford: That is a good idea. I went ahead and renamed the `extra_envs` linked list to just `envs` and it now contains all the defaults and any user added env vars. Wondering if I need to grab the lock in `casterInit` before modifying `self->envs` to add the defaults?

Also moved the argc/argv stuff back to the original functions and re-named the args to `addToReccasterLinkedList` to make it more generic. Does that make sense? should any other code be pulled out of `addToReccasterLinkedList`?

client/castApp/src/caster.c
```diff
@@ -114,9 +116,18 @@ void casterInit(caster_t *self)
...
     ellInit(&self->exclude_patterns);
 
+    /* add default_envs to envs list which can be expanded by the user with addReccasterEnvVars iocsh function */
+    for (i = 0; default_envs[i]; i++) {
+        const size_t arg_len = strlen(default_envs[i]) + 1;
+        string_list_t *new_node = mallocMustSucceed(sizeof(string_list_t) + arg_len, "casterInit");
+        new_node->item_str = (char *)(new_node + 1);
+        memcpy(new_node->item_str, default_envs[i], arg_len);
+        ellAdd(&self->envs, &new_node->node);
+    }
```
simon-ess: Can you not just replace this with a call to
```c
addToReccasterLinkedList(self, num_default_envs, default_envs, self->envs, "casterInit", "default_envs");
```
or something like that?

tynanford: Yes that makes it cleaner. I removed the `itemCount` argument from `addToReccasterLinkedList` since argv and default_envs are both null terminated arrays. It is easy to switch back to use argc and then count the num_default_envs if you think that is better

tynanford: also once the default_env handling was switched to use `addToReccasterLinkedList` , I noticed warnings that there is a duplicate env var `HOSTNAME`. Removed the extra `HOSTNAME` definition in dbcb.c which I don't think will impact anything but thought worth mentioning

tynanford: Chatted with Michael a couple days ago and he said this should use argc or a count instead of relying on the null termination on argv or the default_envs array

client/castApp/src/castinit.c
```diff
@@ -75,20 +75,19 @@ static void casthook(initHookState state)
...
-void addReccasterEnvVars(caster_t* self, int argc, char **argv)
+/* Helper function to add items from iocsh calls to internal linked lists
+ * self is the caster instance
+ * itemCount is the number of items in the items array
+ * items is the array of strings to add to the list
+ * reccastList is the linked list to add the items to
+ * funcName is the name of the IOC shell function being called (for error messages)
+ * itemDesc is string to describe what is being added (for error messages)
+ */
+void addToReccasterLinkedList(caster_t* self, size_t itemCount, const char **items, ELLLIST* reccastList, const char* funcName, const char* itemDesc)
```
simon-ess: One other possible thing to add here is to return a status integer (and same for the shell utils); this could allow for this to be reported back to the shell, etc etc.

simon-ess (approved):

client/castApp/src/castinit.c
tynanford: Ah I didn't know the iocsh functions can do that. Added it and used the `IOCSHFUNCDEF_HAS_USAGE` macro name to check for base compatibility as done in caputlog - https://github.com/epics-modules/caPutLog/blob/master/caPutLogApp/caPutJsonLogShellCommands.cpp#L100

tynanford:
Cleaned up a few things. If this is all good then I can squash some of the commits so it's ready to merge

client/castApp/src/castinit.c
```diff
@@ -205,70 +174,33 @@ static const iocshFuncDef addReccasterEnvVarsFuncDef = {
...
 };
-
-static void addReccasterExcludePatternCallFunc(const iocshArgBuf *args) {
+static void addReccasterExcludePatternCallFunc(const iocshArgBuf *args)
+{
+#ifdef IOCSHFUNCDEF_HAS_USAGE
+    iocshSetError(addReccasterExcludePattern(&thecaster, args[0].aval.ac, args[0].aval.av));
+#else
     addReccasterExcludePattern(&thecaster, args[0].aval.ac, args[0].aval.av);
+#endif
```
simon-ess: I'm not sure I follow this change. Why is this related to `IOCSHFUNCDEF_HAS_USAGE`? I thought that was related (as in the lines above) to adding a bit of help text.

simon-ess: Ah, I see from your comment above what you meant by this. This is a bit confusing IMO, but not obviously something to be done better; `iocshSetError` is from R7.0.3.1 and `IOCSHFUNCDEF_HAS_USAGE` is from R7.0.4, so they are close enough. I sort of would prefer a direct base-version guard on this, but maybe that's a bit silly.

tynanford: Yeah I was looking for examples of how other modules did a base check and saw this. Figured since the base versions are so close it's fine. But I can change to a check on `R7.0.3.1` since that is probably more proper

simon-ess: I'm happy either way. It's a minor comment more directed towards my confusion about what that macro means. I can accept this as is, it's a good change.

tynanford: Figured out the syntax for checking the base version so figured might as well update to that. Squashed a few of the commits together and tested with R3.15.9, R7.0.3, R7.0.3.1, and R7.0.9. All ready from my side

simon-ess (approved):

client/castApp/src/castinit.c
```diff
@@ -205,70 +182,33 @@ static const iocshFuncDef addReccasterEnvVarsFuncDef = {
 };
 static void addReccasterEnvVarsCallFunc(const iocshArgBuf *args)
 {
+#if EPICS_VERSION_INT >= VERSION_INT(7,0,3,1)
+    iocshSetError(addReccasterEnvVars(&thecaster, args[0].aval.ac, args[0].aval.av));
+#else
     addReccasterEnvVars(&thecaster, args[0].aval.ac, args[0].aval.av);
```
mdavidsaver: Duplication of this call could be avoid by defining above:

```c
#if EPICS_VERSION_INT < VERSION_INT(7,0,3,1)
static int iocshSetError(int err) { return err; }
#endif
```

Then `iocshSetError()` will always be available.

tynanford: Ah cool, that is nice

tynanford:
I will plan to merge this tomorrow unless there are objections

mdavidsaver:
Looks ok to me.

image

Added unit tests, and still a net reduction in code lines!

PR #123, opened 2025-10-28
Test restarting channelfinder works ok

jacomago:


server/tests/test_restart.py
```diff
@@ -45,3 +51,27 @@ def test_manual_channels_same_after_restart(
...
+        assert wait_for_sync(cf_client, check_connection_active)
+        channels_end = cf_client.find(name="*")
+        assert len(channels_begin) == len(channels_end)
+        channels_match(channels_begin, channels_end, PROPERTIES_TO_MATCH)
+        shutdown_container(setup_compose, "ioc1-1")
+        assert wait_for_sync(
+            cf_client, lambda cf_client: check_channel_property(cf_client, "IOC1-1:Msg-I", INACTIVE_PROPERTY)
+        )
+        channels_inactive = cf_client.find(property=[("iocName", "IOC1-1")])
+        assert all(INACTIVE_PROPERTY in ch["properties"] for ch in channels_inactive)
```
simon-ess: This seems like two separate tests...

server/tests/test_restart.py
```diff
@@ -45,3 +51,27 @@ def test_manual_channels_same_after_restart(
...
         assert len(channels_begin) == len(channels_end)
         channels_match(channels_begin, channels_end, PROPERTIES_TO_MATCH + ["test_property"])
+
+
+def check_connection_active(cf_client: ChannelFinderClient) -> bool:
+    try:
+        cf_client.find(name="*")
+        return True
+    except Exception:
+        return False
```
anderslindho, line 61: Stylistic, but prefer

```suggestion
    try:
        cf_client.find(name="*")
    except Exception:
        return False
    return True
```

server/tests/test_restart.py
anderslindho, line 106: Prefer when steps have clear setup -> test -> teardown - don't like intermingled asserts, and especially not for setup when you really are testing something else.

Maybe just do 
```suggestion
        wait_for_sync(cf_client, check_connection_active)
        channels_end = cf_client.find(name="*")
        channels_match(channels_begin, channels_end, PROPERTIES_TO_MATCH)
        shutdown_container(setup_compose, "ioc1-1")
        wait_for_sync(
            cf_client, lambda cf_client: check_channel_property(cf_client, "IOC1-1:Msg-I", INACTIVE_PROPERTY)
        )
        channels_inactive = cf_client.find(property=[("iocName", "IOC1-1")])
        
        assert len(channels_begin) == len(channels_end)
        assert all(INACTIVE_PROPERTY in ch["properties"] for ch in channels_inactive)
```
(if those are the two things you want to test here)

does `wait_for_sync` not raise any exceptions itself?

anderslindho (approved):

PR #124, opened 2025-11-04
Refactor cfstore to dataclasses

jacomago:
Large refactor of cfstore

- Adds dataclasses to handle all the intemediary types rather than the usage of dictionaries for everything.
- Swaps to snake case (where not using Twisted stuff)
- Adds typing everywhere I can
- Adds pydoc to every method
- Splits up the bigger methods to be much smaller
- Adds the full data structures to the log messages
- Cleans up exception messages to include more debug information

### DIFF
--- a/server/demo.conf
+++ b/server/demo.conf
@@ -66,7 +66,7 @@
 [cf]
 # cf-store application
 # a space-separated list of infotags to set as CF Properties
-#infotags = archive foo bar blah
+#infotags = archive, foo, bar, blah
 
 # Uncomment line below to turn off the feature to add CA/PVA port info for name server to channelfinder
 #iocConnectionInfo = False
--- a/server/recceiver/cfstore.py
+++ b/server/recceiver/cfstore.py
@@ -222,14 +222,14 @@ class CFChannel:
             "properties": [p.as_dict() for p in self.properties],
         }
 
-    @staticmethod
-    def from_channelfinder_dict(channel_dict: Dict[str, Any]) -> "CFChannel":
+    @classmethod
+    def from_channelfinder_dict(cls, channel_dict: Dict[str, Any]) -> "CFChannel":
         """Create CFChannel from Channelfinder json output.
 
         Args:
             channel_dict: Dictionary representing a channel from Channelfinder.
         """
-        return CFChannel(
+        return cls(
             name=channel_dict.get("name", ""),
             owner=channel_dict.get("owner", ""),
             properties=[CFProperty.from_channelfinder_dict(p) for p in channel_dict.get("properties", [])],
--- a/server/recceiver/cfstore.py
+++ b/server/recceiver/cfstore.py
@@ -1042,6 +1042,14 @@ def create_new_channel(
                 _log.debug("Add new alias: %s from %s", alias, channel_name)
 
 
+class IOCMissingInfoError(Exception):
+    """Raised when an IOC is missing required information."""
+
+    def __init__(self, ioc_info: IocInfo):
+        super().__init__(f"Missing hostName {ioc_info.hostname} or iocName {ioc_info.ioc_name}")
+        self.ioc_info = ioc_info
+
+
 def _update_channelfinder(
     processor: CFProcessor, record_info_by_name: Dict[str, RecordInfo], records_to_delete, ioc_info: IocInfo
 ) -> None:
@@ -1070,7 +1078,7 @@ def _update_channelfinder(
         _log.warning("IOC Env Info %s not found in ioc list: %s", ioc_info, iocs)
 
     if ioc_info.hostname is None or ioc_info.ioc_name is None:
-        raise Exception(f"Missing hostName {ioc_info.hostname} or iocName {ioc_info.ioc_name}")
+        raise IOCMissingInfoError(ioc_info)
 
     if processor.cancelled:
         raise defer.CancelledError(f"Processor cancelled in _update_channelfinder for {ioc_info}")
--- a/server/recceiver/cfstore.py
+++ b/server/recceiver/cfstore.py
@@ -343,9 +343,9 @@ class CFProcessor(service.Service):
         """Stop the CFProcessor service."""
         _log.info("CF_STOP")
         service.Service.stopService(self)
-        return self.lock.run(self._stopServiceWithLock)
+        return self.lock.run(self._stop_service_with_lock)
 
-    def _stopServiceWithLock(self):
+    def _stop_service_with_lock(self):
         """Stop the CFProcessor service with lock held.
 
         If clean_on_stop is enabled, mark all channels as inactive.
--- a/server/recceiver/cfstore.py
+++ b/server/recceiver/cfstore.py
@@ -107,13 +107,29 @@ class PVStatus(enum.Enum):
     Inactive = enum.auto()
 
 
+@dataclass
+class IocInfo:
+    host: str
+    hostname: str
+    ioc_name: str
+    ioc_IP: str
+    owner: str
+    time: str
+    channelcount: int
+    port: int
+
+    @property
+    def ioc_id(self):
+        return self.host + ":" + str(self.port)
+
+
 @implementer(interfaces.IProcessor)
 class CFProcessor(service.Service):
     def __init__(self, name, conf):
         _log.info("CF_INIT {name}".format(name=name))
         self.name = name
-        self.channel_dict = defaultdict(list)
-        self.iocs = dict()
+        self.channel_ioc_ids = defaultdict(list)
+        self.iocs: Dict[str, IocInfo] = dict()
         self.client = None
         self.currentTime = getCurrentTime
         self.lock = DeferredLock()
@@ -268,20 +284,23 @@ class CFProcessor(service.Service):
         {record_id: { "pvName":"recordName",
                 "infoProperties":{propName:value, ...}}}
         """
-
-        host = transaction.source_address.host
-        port = transaction.source_address.port
-        iocName = transaction.client_infos.get("IOCNAME") or transaction.source_address.port
-        hostName = transaction.client_infos.get("HOSTNAME") or transaction.source_address.host
-        owner = (
-            transaction.client_infos.get("ENGINEER")
-            or transaction.client_infos.get("CF_USERNAME")
-            or self.cf_config.username
+        ioc_info = IocInfo(
+            host=transaction.source_address.host,
+            hostname=transaction.client_infos.get("HOSTNAME") or transaction.source_address.host,
+            ioc_name=transaction.client_infos.get("IOCNAME") or str(transaction.source_address.port),
+            ioc_IP=transaction.source_address.host,
+            owner=(
+                transaction.client_infos.get("ENGINEER")
+                or transaction.client_infos.get("CF_USERNAME")
+                or self.cf_config.username
+            ),
+            time=self.currentTime(timezone=self.cf_config.timezone),
+            port=transaction.source_address.port,
+            channelcount=0,
         )
-        time = self.currentTime(timezone=self.cf_config.timezone)
 
         """The unique identifier for a particular IOC"""
-        iocid = host + ":" + str(port)
+        iocid = ioc_info.ioc_id
         _log.debug("transaction: {s}".format(s=repr(transaction)))
 
         recordInfo: Dict[str, RecordInfo] = {}
@@ -302,7 +321,7 @@ class CFProcessor(service.Service):
             if recinfo_wl:
                 for infotag in recinfo_wl:
                     recordInfo[record_id].infoProperties.append(
-                        CFProperty(infotag, owner, record_infos_to_add[infotag])
+                        CFProperty(infotag, ioc_info.owner, record_infos_to_add[infotag])
                     )
 
         for record_id, alias in transaction.aliases.items():
@@ -319,13 +338,13 @@ class CFProcessor(service.Service):
             for epics_env_var_name, cf_prop_name in self.env_vars.items():
                 if transaction.client_infos.get(epics_env_var_name) is not None:
                     recordInfo[record_id].infoProperties.append(
-                        CFProperty(cf_prop_name, owner, transaction.client_infos.get(epics_env_var_name))
+                        CFProperty(cf_prop_name, ioc_info.owner, transaction.client_infos.get(epics_env_var_name))
                     )
                 else:
                     _log.debug(
                         "EPICS environment var %s listed in environment_vars setting list not found in this IOC: %s",
                         epics_env_var_name,
-                        iocName,
+                        ioc_info.ioc_name,
                     )
 
         records_to_delete = list(transaction.records_to_delete)
@@ -342,56 +361,38 @@ class CFProcessor(service.Service):
 
         if transaction.initial:
             """Add IOC to source list """
-            self.iocs[iocid] = {
-                "iocname": iocName,
-                "hostname": hostName,
-                "iocIP": host,
-                "owner": owner,
-                "time": time,
-                "channelcount": 0,
-            }
+            self.iocs[iocid] = ioc_info
         if not transaction.connected:
-            records_to_delete.extend(self.channel_dict.keys())
+            records_to_delete.extend(self.channel_ioc_ids.keys())
         for record_name in recordInfoByName.keys():
-            self.channel_dict[record_name].append(iocid)
-            self.iocs[iocid]["channelcount"] += 1
+            self.channel_ioc_ids[record_name].append(iocid)
+            self.iocs[iocid].channelcount += 1
             """In case, alias exists"""
             if self.cf_config.alias_enabled:
                 if record_name in recordInfoByName:
                     for alias in recordInfoByName[record_name].aliases:
-                        self.channel_dict[alias].append(iocid)  # add iocname to pvName in dict
-                        self.iocs[iocid]["channelcount"] += 1
+                        self.channel_ioc_ids[alias].append(iocid)  # add iocname to pvName in dict
+                        self.iocs[iocid].channelcount += 1
         for record_name in records_to_delete:
-            if iocid in self.channel_dict[record_name]:
+            if iocid in self.channel_ioc_ids[record_name]:
                 self.remove_channel(record_name, iocid)
                 """In case, alias exists"""
                 if self.cf_config.alias_enabled:
                     if record_name in recordInfoByName:
                         for alias in recordInfoByName[record_name].aliases:
                             self.remove_channel(alias, iocid)
-        poll(
-            __updateCF__,
-            self,
-            recordInfoByName,
-            records_to_delete,
-            hostName,
-            iocName,
-            host,
-            iocid,
-            owner,
-            time,
-        )
+        poll(__updateCF__, self, recordInfoByName, records_to_delete, ioc_info)
 
     def remove_channel(self, recordName, iocid):
-        self.channel_dict[recordName].remove(iocid)
+        self.channel_ioc_ids[recordName].remove(iocid)
         if iocid in self.iocs:
-            self.iocs[iocid]["channelcount"] -= 1
-        if self.iocs[iocid]["channelcount"] == 0:
+            self.iocs[iocid].channelcount -= 1
+        if self.iocs[iocid].channelcount == 0:
             self.iocs.pop(iocid, None)
-        elif self.iocs[iocid]["channelcount"] < 0:
+        elif self.iocs[iocid].channelcount < 0:
             _log.error("Channel count negative: {s}", s=iocid)
-        if len(self.channel_dict[recordName]) <= 0:  # case: channel has no more iocs
-            del self.channel_dict[recordName]
+        if len(self.channel_ioc_ids[recordName]) <= 0:  # case: channel has no more iocs
+            del self.channel_ioc_ids[recordName]
 
     def clean_service(self):
         """
@@ -478,41 +479,26 @@ def create_time_property(owner: str, time: str) -> CFProperty:
     return CFProperty(CFPropertyName.time.name, owner, time)
 
 
-def __updateCF__(
-    processor: CFProcessor,
-    recordInfoByName: Dict[str, RecordInfo],
-    records_to_delete,
-    hostName,
-    iocName,
-    iocIP,
-    iocid,
-    owner,
-    iocTime,
-):
-    _log.info("CF Update IOC: {iocid}".format(iocid=iocid))
+def __updateCF__(processor: CFProcessor, recordInfoByName: Dict[str, RecordInfo], records_to_delete, ioc_info: IocInfo):
+    _log.info("CF Update IOC: {iocid}".format(iocid=ioc_info.ioc_id))
     _log.debug(
         "CF Update IOC: {iocid} recordInfoByName {recordInfoByName}".format(
-            iocid=iocid, recordInfoByName=recordInfoByName
+            iocid=ioc_info.ioc_id, recordInfoByName=recordInfoByName
         )
     )
     # Consider making this function a class methed then 'processor' simply becomes 'self'
     client = processor.client
-    channels_dict = processor.channel_dict
+    channel_ioc_ids = processor.channel_ioc_ids
     iocs = processor.iocs
     cf_config = processor.cf_config
     recceiverid = processor.cf_config.recceiver_id
     new_channels = set(recordInfoByName.keys())
+    iocid = ioc_info.ioc_id
 
-    if iocid in iocs:
-        hostName = iocs[iocid]["hostname"]
-        iocName = iocs[iocid]["iocname"]
-        owner = iocs[iocid]["owner"]
-        iocTime = iocs[iocid]["time"]
-        iocIP = iocs[iocid]["iocIP"]
-    else:
+    if iocid not in iocs:
         _log.warning("IOC Env Info not found: {iocid}".format(iocid=iocid))
 
-    if hostName is None or iocName is None:
+    if ioc_info.hostname is None or ioc_info.ioc_name is None:
         raise Exception("missing hostName or iocName")
 
     if processor.cancelled:
@@ -540,10 +526,10 @@ def __updateCF__(
                 len(new_channels) == 0 or cf_channel["name"] in records_to_delete
             ):  # case: empty commit/del, remove all reference to ioc
                 _log.debug("Channel {s} exists in Channelfinder not in new_channels".format(s=cf_channel["name"]))
-                if cf_channel["name"] in channels_dict:
-                    cf_channel["owner"] = iocs[channels_dict[cf_channel["name"]][-1]]["owner"]
+                if cf_channel["name"] in channel_ioc_ids:
+                    cf_channel["owner"] = iocs[channel_ioc_ids[cf_channel["name"]][-1]].owner
                     cf_channel["properties"] = __merge_property_lists(
-                        create_default_properties(owner, iocTime, recceiverid, channels_dict, iocs, cf_channel),
+                        create_default_properties(ioc_info, recceiverid, channel_ioc_ids, iocs, cf_channel),
                         cf_channel,
                         processor.managed_properties,
                     )
@@ -551,7 +537,7 @@ def __updateCF__(
                         cf_channel["properties"] = __merge_property_lists(
                             cf_channel["properties"].append(
                                 create_recordType_property(
-                                    owner, iocs[channels_dict[cf_channel["name"]][-1]]["recordType"]
+                                    ioc_info.owner, iocs[channel_ioc_ids[cf_channel["name"]][-1]]["recordType"]
                                 )
                             ),
                             cf_channel,
@@ -565,14 +551,13 @@ def __updateCF__(
                             for alias_name in recordInfoByName[cf_channel["name"]].aliases:
                                 # TODO Remove? This code couldn't have been working....
                                 alias_channel = {"name": alias_name, "properties": [], "owner": ""}
-                                if alias_name in channels_dict:
-                                    alias_channel["owner"] = iocs[channels_dict[alias_name][-1]]["owner"]
+                                if alias_name in channel_ioc_ids:
+                                    alias_channel["owner"] = iocs[channel_ioc_ids[alias_name][-1]].owner
                                     alias_channel["properties"] = __merge_property_lists(
                                         create_default_properties(
-                                            owner,
-                                            iocTime,
+                                            ioc_info,
                                             recceiverid,
-                                            channels_dict,
+                                            channel_ioc_ids,
                                             iocs,
                                             cf_channel,
                                         ),
@@ -583,8 +568,8 @@ def __updateCF__(
                                         cf_channel["properties"] = __merge_property_lists(
                                             cf_channel["properties"].append(
                                                 create_recordType_property(
-                                                    owner,
-                                                    iocs[channels_dict[alias_name][-1]]["recordType"],
+                                                    ioc_info.owner,
+                                                    iocs[channel_ioc_ids[alias_name][-1]]["recordType"],
                                                 )
                                             ),
                                             cf_channel,
@@ -597,8 +582,8 @@ def __updateCF__(
                     """Orphan the channel : mark as inactive, keep the old hostName and iocName"""
                     cf_channel["properties"] = __merge_property_lists(
                         [
-                            create_inactive_property(owner),
-                            create_time_property(owner, iocTime),
+                            create_inactive_property(ioc_info.owner),
+                            create_time_property(ioc_info.owner, ioc_info.time),
                         ],
                         cf_channel,
                     )
@@ -611,8 +596,8 @@ def __updateCF__(
                                 alias_channel = {"name": alias_name, "properties": [], "owner": ""}
                                 alias_channel["properties"] = __merge_property_lists(
                                     [
-                                        create_inactive_property(owner),
-                                        create_time_property(owner, iocTime),
+                                        create_inactive_property(ioc_info.owner),
+                                        create_time_property(ioc_info.owner, ioc_info.time),
                                     ],
                                     alias_channel,
                                 )
@@ -631,8 +616,8 @@ def __updateCF__(
                     )
                     cf_channel["properties"] = __merge_property_lists(
                         [
-                            create_active_property(owner),
-                            create_time_property(owner, iocTime),
+                            create_active_property(ioc_info.owner),
+                            create_time_property(ioc_info.owner, ioc_info.time),
                         ],
                         cf_channel,
                         processor.managed_properties,
@@ -650,8 +635,8 @@ def __updateCF__(
                                     alias_channel = {"name": alias_name, "properties": [], "owner": ""}
                                     alias_channel["properties"] = __merge_property_lists(
                                         [
-                                            create_active_property(owner),
-                                            create_time_property(owner, iocTime),
+                                            create_active_property(ioc_info.owner),
+                                            create_time_property(ioc_info.owner, ioc_info.time),
                                         ],
                                         alias_channel,
                                         processor.managed_properties,
@@ -662,10 +647,10 @@ def __updateCF__(
                                     """alias exists but not part of old list"""
                                     aprops = __merge_property_lists(
                                         [
-                                            create_active_property(owner),
-                                            create_time_property(owner, iocTime),
+                                            create_active_property(ioc_info.owner),
+                                            create_time_property(ioc_info.owner, ioc_info.time),
                                             create_alias_property(
-                                                owner,
+                                                ioc_info.owner,
                                                 cf_channel["name"],
                                             ),
                                         ],
@@ -675,7 +660,7 @@ def __updateCF__(
                                     channels.append(
                                         create_channel(
                                             alias_name,
-                                            owner,
+                                            ioc_info.owner,
                                             aprops,
                                         )
                                     )
@@ -714,13 +699,21 @@ def __updateCF__(
             raise defer.CancelledError()
 
     for channel_name in new_channels:
-        newProps = create_properties(owner, iocTime, recceiverid, hostName, iocName, iocIP, iocid)
+        newProps = create_properties(
+            ioc_info.owner,
+            ioc_info.time,
+            recceiverid,
+            ioc_info.hostname,
+            ioc_info.ioc_name,
+            ioc_info.ioc_IP,
+            ioc_info.ioc_id,
+        )
         if (
             cf_config.record_type_enabled
             and channel_name in recordInfoByName
             and recordInfoByName[channel_name].recordType
         ):
-            newProps.append(create_recordType_property(owner, recordInfoByName[channel_name].recordType))
+            newProps.append(create_recordType_property(ioc_info.owner, recordInfoByName[channel_name].recordType))
         if channel_name in recordInfoByName:
             newProps = newProps + recordInfoByName[channel_name].infoProperties
 
@@ -740,7 +733,7 @@ def __updateCF__(
             """in case, alias exists, update their properties too"""
             if cf_config.alias_enabled:
                 if channel_name in recordInfoByName:
-                    alProps = [create_alias_property(owner, channel_name)]
+                    alProps = [create_alias_property(ioc_info.owner, channel_name)]
                     for p in newProps:
                         alProps.append(p)
                     for alias_name in recordInfoByName[channel_name].aliases:
@@ -753,22 +746,24 @@ def __updateCF__(
                             )
                             channels.append(ach)
                         else:
-                            channels.append(create_channel(alias_name, owner, alProps))
+                            channels.append(create_channel(alias_name, ioc_info.owner, alProps))
                         _log.debug("Add existing alias with different IOC: {s}".format(s=channels[-1]))
 
         else:
             """New channel"""
-            channels.append({"name": channel_name, "owner": owner, "properties": newProps})
+            channels.append({"name": channel_name, "owner": ioc_info.owner, "properties": newProps})
             _log.debug("Add new channel: {s}".format(s=channels[-1]))
             if cf_config.alias_enabled:
                 if channel_name in recordInfoByName:
-                    alProps = [create_alias_property(owner, channel_name)]
+                    alProps = [create_alias_property(ioc_info.owner, channel_name)]
                     for p in newProps:
                         alProps.append(p)
                     for alias in recordInfoByName[channel_name].aliases:
-                        channels.append({"name": alias, "owner": owner, "properties": alProps})
+                        channels.append({"name": alias, "owner": ioc_info.owner, "properties": alProps})
                         _log.debug("Add new alias: {s}".format(s=channels[-1]))
-    _log.info("Total channels to update: {nChannels} {iocName}".format(nChannels=len(channels), iocName=iocName))
+    _log.info(
+        "Total channels to update: {nChannels} {iocName}".format(nChannels=len(channels), iocName=ioc_info.ioc_name)
+    )
     if len(channels) != 0:
         cf_set_chunked(client, channels, cf_config.cf_query_limit)
     else:
@@ -788,7 +783,7 @@ def cf_set_chunked(client, channels, chunk_size=10000):
         client.set(channels=chunk)
 
 
-def create_properties(owner, iocTime, recceiverid, hostName, iocName, iocIP, iocid):
+def create_properties(owner: str, iocTime: str, recceiverid: str, hostName: str, iocName: str, iocIP: str, iocid: str):
     return [
         CFProperty(CFPropertyName.hostName.name, owner, hostName),
         CFProperty(CFPropertyName.iocName.name, owner, iocName),
@@ -800,15 +795,19 @@ def create_properties(owner, iocTime, recceiverid, hostName, iocName, iocIP, ioc
     ]
 
 
-def create_default_properties(owner, iocTime, recceiverid, channels_dict, iocs, cf_channel):
+def create_default_properties(
+    ioc_info: IocInfo, recceiverid: str, channels_iocs: Dict[str, List[str]], iocs: Dict[str, IocInfo], cf_channel
+):
+    channel_name = cf_channel["name"]
+    last_ioc_info = iocs[channels_iocs[channel_name][-1]]
     return create_properties(
-        owner,
-        iocTime,
+        ioc_info.owner,
+        ioc_info.time,
         recceiverid,
-        iocs[channels_dict[cf_channel["name"]][-1]]["hostname"],
-        iocs[channels_dict[cf_channel["name"]][-1]]["iocname"],
-        iocs[channels_dict[cf_channel["name"]][-1]]["iocIP"],
-        channels_dict[cf_channel["name"]][-1],
+        last_ioc_info.hostname,
+        last_ioc_info.ioc_name,
+        last_ioc_info.ioc_IP,
+        last_ioc_info.ioc_id,
     )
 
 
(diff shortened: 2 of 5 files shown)


server/recceiver/cfstore.py
```diff
@@ -103,25 +135,25 @@ def _startServiceWithLock(self):
...
                     self.env_vars["RSRV_SERVER_PORT"] = "caPort"
                     self.env_vars["PVAS_SERVER_PORT"] = "pvaPort"
                     required_properties.add("caPort")
                     required_properties.add("pvaPort")
-                infotags_whitelist = self.conf.get("infotags", list())
+                infotags_whitelist = self.cf_config.infotags
                 if infotags_whitelist:
                     record_property_names_list = [s.strip(", ") for s in infotags_whitelist.split()]
                 else:
                     record_property_names_list = []
```
simon-ess: Do you need this conditional block now? The default is an empty string, so 
```py
record_property_names_list = [s.strip(", ") for s in infotags_whitelist.split()]
```
should return an empty list in that case.

server/recceiver/cfstore.py
```diff
@@ -39,16 +37,50 @@
...
+    environment_vars: str = ""
+    infotags: str = ""
+    iocConnectionInfo: bool = True
+    recordDesc: bool = False
+    cleanOnStart: bool = True
+    cleanOnStop: bool = True
+    username: str = "cfstore"
+    recceiverId: str = RECCEIVERID_DEFAULT
+    timezone: str = ""
+    findSizeLimit: int = 10000
```
simon-ess: cAmElCaSe vs s_n_a_k_e_c_a_s_e?

server/recceiver/cfstore.py
```diff
@@ -735,7 +738,10 @@ def __updateCF__(
 
 def cf_set_chunked(client, channels, chunk_size=10000):
     for i in range(0, len(channels), chunk_size):
-        chunk = channels[i : i + chunk_size]
+        chunk = [
+            {"name": ch["name"], "owner": ch["owner"], "properties": ch["properties"].as_dict()}
+            for ch in channels[i : i + chunk_size]
+        ]
         client.set(channels=chunk)
```
simon-ess: Some type hints here would be nice. It's not clear what `channels` is.

server/recceiver/cfstore.py
```diff
@@ -110,7 +117,7 @@ def _startServiceWithLock(self):
             """
             self.client = ChannelFinderClient()
             try:
-                cf_properties = [cf_property["name"] for cf_property in self.client.getAllProperties()]
+                cf_properties = {cf_property["name"] for cf_property in self.client.getAllProperties()}
```
simon-ess, line 292: I see that this is used as a `set` below; if you're making this change, maybe remove the `set()`s below?

simon-ess: Also, you're creating a CFProperty data class here but here you're reading things as a dictionary. This commit is a bit confusing.

server/recceiver/cfstore.py
```diff
@@ -74,7 +65,10 @@ def load_from_conf(conf) -> "CFConfig":
 class CFProperty:
     name: str
     owner: str
-    value: str
+    value: Optional[str] = None
+
+    def as_dict(self) -> Dict[str, str]:
+        return {"name": self.name, "owner": self.owner, "value": str(self.value)}
```
simon-ess: This should probably be merged into the previous commit as you are using it there...

server/recceiver/interfaces.py
```diff
@@ -21,6 +24,24 @@ class ITransaction(Interface):
...
+    source_address: SourceAddress
+    client_infos: Dict[str, str]
+    records_to_add: Dict[str, Tuple[str, str]]
+    records_to_delete: Set[str]
+    record_infos_to_add: Dict[str, Dict[str, str]]
+    aliases: Dict[str, List[str]]
+    initial: bool
+    connected: bool
+
+
```
simon-ess, line 44: This commit seems odd on its own. You create two dataclasses, one of which is not used and one of which is only used as a type hint. Should this be its own commit?

server/recceiver/cfstore.py
```diff
@@ -251,11 +259,11 @@ def _commitWithThread(self, transaction: CommitTransaction):
...
         _log.debug("transaction: {s}".format(s=repr(transaction)))
 
-        recordInfo = {}
+        recordInfo: Dict[str, RecordInfo] = {}
         for record_id, (record_name, record_type) in transaction.records_to_add.items():
-            recordInfo[record_id] = {"pvName": record_name}
+            recordInfo[record_id] = RecordInfo(pvName=record_name, recordType=None, infoProperties=[], aliases=[])
             if self.cf_config.recordType:
-                recordInfo[record_id]["recordType"] = record_type
+                recordInfo[record_id].recordType = record_type
```
simon-ess: This feels ugly somehow, but I'm not sure what to do about it.

server/recceiver/cfstore.py
```diff
@@ -267,9 +275,8 @@ def _commitWithThread(self, transaction: CommitTransaction):
                 continue
             recinfo_wl = [p for p in self.record_property_names_list if p in record_infos_to_add.keys()]
             if recinfo_wl:
-                recordInfo[record_id]["infoProperties"] = list()
                 for infotag in recinfo_wl:
-                    recordInfo[record_id]["infoProperties"].append(
+                    recordInfo[record_id].infoProperties.append(
                         create_property(owner, infotag, record_infos_to_add[infotag])
                     )
```
simon-ess, line 450: This conditional can be removed now since the "infoProperties" field already exists in the dataclass.

server/recceiver/cfstore.py
```diff
@@ -329,17 +332,17 @@ def _commitWithThread(self, transaction: CommitTransaction):
             self.iocs[iocid]["channelcount"] += 1
             """In case, alias exists"""
             if self.cf_config.alias:
-                if record_name in recordInfoByName and "aliases" in recordInfoByName[record_name]:
-                    for alias in recordInfoByName[record_name]["aliases"]:
+                if record_name in recordInfoByName:
+                    for alias in recordInfoByName[record_name].aliases:
```
simon-ess: SO MUCH CLEANER

server/recceiver/cfstore.py
```diff
@@ -525,11 +528,13 @@ def __updateCF__(
...
                     if cf_config.alias:
-                        if cf_channel["name"] in recordInfoByName and "aliases" in recordInfoByName[cf_channel["name"]]:
-                            for alias in recordInfoByName[cf_channel["name"]]["aliases"]:
-                                if alias["name"] in channels_dict:
-                                    alias["owner"] = iocs[channels_dict[alias["name"]][-1]]["owner"]
-                                    alias["properties"] = __merge_property_lists(
+                        if cf_channel["name"] in recordInfoByName:
+                            for alias_name in recordInfoByName[cf_channel["name"]].aliases:
+                                # TODO Remove? This code couldn't have been working....
+                                alias_channel = {"name": alias_name, "properties": [], "owner": ""}
```
simon-ess: Don't we have a dataclass for this yet?

server/recceiver/cfstore.py
```diff
@@ -525,11 +528,13 @@ def __updateCF__(
...
+                                if alias_name in channels_dict:
```
simon-ess: Is this the part that "couldn't have been workign"? That is, the switch from whatever the `alias` thingy was before (evidently a dictionary) to a string?

server/recceiver/cfstore.py
```diff
@@ -671,10 +678,10 @@ def __updateCF__(
 
     for channel_name in new_channels:
         newProps = create_properties(owner, iocTime, recceiverid, hostName, iocName, iocIP, iocid)
-        if cf_config.recordType:
-            newProps.append(create_recordType_property(owner, recordInfoByName[channel_name]["recordType"]))
-        if channel_name in recordInfoByName and "infoProperties" in recordInfoByName[channel_name]:
-            newProps = newProps + recordInfoByName[channel_name]["infoProperties"]
+        if cf_config.recordType and channel_name in recordInfoByName and recordInfoByName[channel_name].recordType:
```
simon-ess: Is this an edge case that you have found?

server/recceiver/cfstore.py
```diff
@@ -79,6 +79,22 @@ class RecordInfo:
...
+    iocIP = enum.auto()
+    pvStatus = enum.auto()
+    time = enum.auto()
+    recceiverID = enum.auto()
+    alias = enum.auto()
+    recordType = enum.auto()
+    recordDesc = enum.auto()
+    caPort = enum.auto()
+    pvaPort = enum.auto()
+    time = enum.auto()
```
simon-ess: I can see the advantage here of getting rid of magic strings, but aren't properties somewhat mutable? That is, people can add their own properties above and beyond this default list.

server/recceiver/cfstore.py
```diff
@@ -338,25 +355,18 @@ def _commitWithThread(self, transaction: CommitTransaction):
 
         if transaction.initial:
             """Add IOC to source list """
```
simon-ess: I am finding these """ comments very distracting. Who wrote these????

server/recceiver/cfstore.py
```diff
@@ -478,41 +477,26 @@ def create_time_property(owner: str, time: str) -> CFProperty:
...
     _log.debug(
         "CF Update IOC: {iocid} recordInfoByName {recordInfoByName}".format(
-            iocid=iocid, recordInfoByName=recordInfoByName
+            iocid=ioc_info.ioc_id, recordInfoByName=recordInfoByName
         )
     )
     # Consider making this function a class methed then 'processor' simply becomes 'self'
     client = processor.client
-    channels_dict = processor.channel_dict
+    channels_iocs = processor.channel_dict
```
simon-ess: I don't get this name change.

server/recceiver/cfstore.py
```diff
@@ -531,18 +515,18 @@ def __updateCF__(
                 len(new_channels) == 0 or cf_channel["name"] in records_to_delete
             ):  # case: empty commit/del, remove all reference to ioc
                 _log.debug("Channel {s} exists in Channelfinder not in new_channels".format(s=cf_channel["name"]))
-                if cf_channel["name"] in channels_dict:
-                    cf_channel["owner"] = iocs[channels_dict[cf_channel["name"]][-1]]["owner"]
+                if cf_channel["name"] in channels_iocs:
+                    cf_channel["owner"] = iocs[channels_iocs[cf_channel["name"]][-1]].owner
```
simon-ess: This is an improvement, but UGH

server/recceiver/cfstore.py
```diff
@@ -531,18 +515,18 @@ def __updateCF__(
...
+                        create_default_properties(ioc_info, recceiverid, channels_iocs, iocs, cf_channel),
                         cf_channel,
                         processor.managed_properties,
                     )
                     if cf_config.recordType:
                         cf_channel["properties"] = __merge_property_lists(
                             cf_channel["properties"].append(
                                 create_recordType_property(
-                                    owner, iocs[channels_dict[cf_channel["name"]][-1]]["recordType"]
+                                    ioc_info.owner, iocs[channels_iocs[cf_channel["name"]][-1]]["recordType"]
```
simon-ess: Should this be a `.recordType` here? We use `.owner` above on the same object...

(These objects are very confusing to follow... this is clearly going to be an improvement but ugh)

server/recceiver/cfstore.py
```diff
@@ -782,15 +775,19 @@ def create_properties(owner, iocTime, recceiverid, hostName, iocName, iocIP, ioc
     ]
 
 
-def create_default_properties(owner, iocTime, recceiverid, channels_dict, iocs, cf_channel):
+def create_default_properties(
+    ioc_info: IocInfo, recceiverid: str, channels_iocs: Dict[str, List[str]], iocs: Dict[str, IocInfo], cf_channel
+):
+    channel_name = cf_channel["name"]
+    last_ioc_info = iocs[channels_iocs[channel_name][-1]]
```
simon-ess, line 1211: Maybe use a similar definition above?

server/recceiver/cfstore.py
```diff
@@ -515,83 +514,6 @@ def __updateCF__(processor: CFProcessor, recordInfoByName: Dict[str, RecordInfo]
                 len(new_channels) == 0 or cf_channel["name"] in records_to_delete
             ):  # case: empty commit/del, remove all reference to ioc
                 _log.debug("Channel {s} exists in Channelfinder not in new_channels".format(s=cf_channel["name"]))
```
simon-ess: I am uncomfortable with the ? next to impossible code in the commit message. Can you clarify why this is impossible to hit code?

server/recceiver/cfstore.py
```diff
@@ -493,38 +489,34 @@ def __updateCF__(processor: CFProcessor, recordInfoByName: Dict[str, RecordInfo]
     iocid = ioc_info.ioc_id
 
     if iocid not in iocs:
-        _log.warning("IOC Env Info not found: {iocid}".format(iocid=iocid))
+        _log.warning("IOC Env Info %s not found in ioc list: %s", ioc_info, iocs)
```
simon-ess, line 1078: Is it possible that this will result in giant logs? If we are printing every IOC here...

simon-ess (commented):
Overall this is a clear improvement!

server/recceiver/cfstore.py
```diff
@@ -39,16 +37,50 @@
...
+    iocConnectionInfo: bool = True
+    recordDesc: bool = False
+    cleanOnStart: bool = True
+    cleanOnStop: bool = True
+    username: str = "cfstore"
+    recceiverId: str = RECCEIVERID_DEFAULT
+    timezone: str = ""
+    findSizeLimit: int = 10000
+
+    @staticmethod
```
anderslindho: Shouldn't this be a class method?

I'd also suggest changing the name to drop "load"; maybe just `from_config_file` 🔧

server/recceiver/cfstore.py
```diff
@@ -525,11 +528,13 @@ def __updateCF__(
...
                     """In case alias exist, also delete them"""
                     if cf_config.alias:
-                        if cf_channel["name"] in recordInfoByName and "aliases" in recordInfoByName[cf_channel["name"]]:
-                            for alias in recordInfoByName[cf_channel["name"]]["aliases"]:
-                                if alias["name"] in channels_dict:
-                                    alias["owner"] = iocs[channels_dict[alias["name"]][-1]]["owner"]
-                                    alias["properties"] = __merge_property_lists(
+                        if cf_channel["name"] in recordInfoByName:
+                            for alias_name in recordInfoByName[cf_channel["name"]].aliases:
+                                # TODO Remove? This code couldn't have been working....
```
anderslindho: Is there a task or GH issue to track this?

server/recceiver/cfstore.py
```diff
@@ -604,20 +610,21 @@ def __updateCF__(
...
+                                    alias_channel = {"name": alias_name, "properties": [], "owner": ""}
+                                    alias_channel["properties"] = __merge_property_lists(
                                         [
                                             create_active_property(owner),
                                             create_time_property(owner, iocTime),
                                         ],
-                                        alias,
+                                        alias_channel,
                                         processor.managed_properties,
                                     )
```
anderslindho: Why not do this in the above row when you're declaring the dict? 🔧

server/recceiver/cfstore.py
```diff
@@ -79,6 +79,22 @@ class RecordInfo:
     aliases: List[str] = field(default_factory=list)
 
 
+class CFPropertyName(enum.Enum):
+    hostName = enum.auto()
```
anderslindho: IMO just import `Enum` and `auto` and skip the namespaces here 🔧

server/recceiver/cfstore.py
```diff
@@ -100,13 +100,28 @@ class PVStatus(enum.Enum):
     Inactive = enum.auto()
 
 
+@dataclass
+class IocInfo:
```
anderslindho, line 191: Preferential, but maybe do `IOCInfo` 🔧

server/recceiver/cfstore.py
```diff
@@ -100,13 +100,28 @@ class PVStatus(enum.Enum):
...
+    iocname: str
+    iocIP: str
+    owner: str
+    time: str
+    channelcount: int
+    port: int
+
+    @property
+    def ioc_id(self):
+        return self.hostname + ":" + self.port
```
anderslindho: Maybe drop the setter?

server/recceiver/cfstore.py
```diff
@@ -736,22 +727,24 @@ def __updateCF__(
...
                     for p in newProps:
                         alProps.append(p)
                     for alias in recordInfoByName[channel_name].aliases:
-                        channels.append({"name": alias, "owner": owner, "properties": alProps})
+                        channels.append({"name": alias, "owner": ioc_info.owner, "properties": alProps})
                         _log.debug("Add new alias: {s}".format(s=channels[-1]))
-    _log.info("Total channels to update: {nChannels} {iocName}".format(nChannels=len(channels), iocName=iocName))
+    _log.info(
+        "Total channels to update: {nChannels} {iocName}".format(nChannels=len(channels), iocName=ioc_info.iocname)
+    )
```
anderslindho: Convert to C-style strings for performance since you're touching this anyway? (for the ones you do touch) 🔧

server/recceiver/cfstore.py
anderslindho, line 1078: Can we _please_ improve this msg/phrasing while we're at it? I _hate_ "IOC Env Info"

server/recceiver/cfstore.py
```diff
@@ -493,38 +489,34 @@ def __updateCF__(processor: CFProcessor, recordInfoByName: Dict[str, RecordInfo]
     iocid = ioc_info.ioc_id
 
     if iocid not in iocs:
-        _log.warning("IOC Env Info not found: {iocid}".format(iocid=iocid))
+        _log.warning("IOC Env Info %s not found in ioc list: %s", ioc_info, iocs)
 
     if ioc_info.hostname is None or ioc_info.iocname is None:
-        raise Exception("missing hostName or iocName")
+        raise Exception(f"Missing hostName {ioc_info.hostname} or iocName {ioc_info.iocname}")
```
anderslindho: It's f-strings so you can just hello `{ioc_info.hostname=}` 🔧 

**Edit:** Urgh, we're on an ancient python version, so maybe not

server/recceiver/cfstore.py
```diff
@@ -585,9 +506,96 @@ def __updateCF__(processor: CFProcessor, recordInfoByName: Dict[str, RecordInfo]
...
+    records_to_delete: List[str],
+    ioc_info: IocInfo,
+    managed_properties: Set[str],
+    channels: List[CFChannel],
+    alias_enabled: bool,
+    recordInfoByName: Dict[str, RecordInfo],
+):
+    for cf_channel in old_channels:
+        if (
+            len(new_channels) == 0 or cf_channel.name in records_to_delete
```
anderslindho: No reason not to do `not new_channels or` ...

server/recceiver/cfstore.py
```diff
@@ -645,6 +653,57 @@ def __updateCF__(processor: CFProcessor, recordInfoByName: Dict[str, RecordInfo]
                     for alias in recordInfoByName[channel_name].aliases:
                         channels.append(CFChannel(alias, ioc_info.owner, alProps))
                         _log.debug("Add new alias: %s from %s", alias, channel_name)
+    return alias
+
+
+def __updateCF__(processor: CFProcessor, recordInfoByName: Dict[str, RecordInfo], records_to_delete, ioc_info: IocInfo):
```
anderslindho: We should avoid imitating dunder methods - just do `__updateCF()`

server/recceiver/cfstore.py
jacomago: Eh, I prefer the namespace for enum.

server/recceiver/cfstore.py
jacomago: setter?

server/recceiver/interfaces.py
jacomago: This is only for type hinting, in reality the object passing is an ITransaction, but then you don't get the full type info. And I have a feeling it might mess up other things to touch it.

server/recceiver/cfstore.py
anderslindho: https://docs.python.org/3/library/functions.html#property.setter

server/recceiver/cfstore.py
anderslindho: FWIW they are not comments, just random unbound string literals 🤷

simon-ess: Yes, but they are  clearly used as comments which is super annoying.

server/recceiver/cfstore.py
jacomago: Yes, this is just for default properties. Not for instance the environment properties or the info tag properties.

server/recceiver/cfstore.py
jacomago: So channels_dict is actually a dictionary mapping a channel to a list of iocs that the channel is associated to. Tried to make clearer.

server/recceiver/cfstore.py
anderslindho: This comment seems to still apply

server/recceiver/cfstore.py
```diff
@@ -103,25 +136,22 @@ def _startServiceWithLock(self):
...
                     required_properties.add("caPort")
                     required_properties.add("pvaPort")
-                infotags_whitelist = self.conf.get("infotags", list())
-                if infotags_whitelist:
-                    record_property_names_list = [s.strip(", ") for s in infotags_whitelist.split()]
-                else:
-                    record_property_names_list = []
-                if self.conf.getboolean("recordDesc"):
+
+                record_property_names_list = [s.strip(", ") for s in self.cf_config.info_tags.split()]
```
anderslindho: Comment above this row showing example entries would be nice 🔧

server/recceiver/cfstore.py
```diff
@@ -71,6 +71,16 @@ def from_config_adapter(conf: ConfigAdapter) -> "CFConfig":
...
 
 
+@dataclass
+class CFProperty:
+    name: str
+    owner: str
+    value: Optional[str] = None
+
+    def as_dict(self) -> Dict[str, str]:
+        return {"name": self.name, "owner": self.owner, "value": str(self.value)}
```
anderslindho: Given `value: Optional` I'd say this should be
```python
    def as_dict(self) -> Dict[str, Optional[str]]:
        return {"name": self.name, "owner": self.owner, "value": str(self.value)}
```
and you're probably better off doing `"value": self.value or ""` (to avoid the string `"None"`)

server/recceiver/cfstore.py
```diff
@@ -430,7 +420,7 @@ def create_channel(name: str, owner: str, properties: List[CFProperty]):
     }
 
 
-def create_property(owner: str, name: str, value: str) -> CFProperty:
+def create_property(owner: str, name: str, value: Optional[str]) -> CFProperty:
```
anderslindho: Does this not have any implications on the `CFProperty` ctor? There is no associated change to that class in this commit.

server/recceiver/cfstore.py
```diff
@@ -474,93 +472,114 @@ def create_time_property(owner: str, time: str) -> CFProperty:
...
+) -> Dict[str, CFChannel]:
+    """A dictionary representing the current channelfinder information associated with the pvNames"""
+    existingChannels = {}
 
-    if ioc_info.hostname is None or ioc_info.iocname is None:
-        raise Exception(f"Missing hostName {ioc_info.hostname} or iocName {ioc_info.iocname}")
+    """
+    The list of pv's is searched keeping in mind the limitations on the URL length
+    The search is split into groups to ensure that the size does not exceed 600 characters
+    """
```
anderslindho, line 788: Let's not add further throwaway string literals here - use a comment instead please

server/recceiver/cfstore.py
```diff
@@ -474,93 +472,114 @@ def create_time_property(owner: str, time: str) -> CFProperty:
...
+    searchStrings = []
+    searchString = ""
+    for channel_name in new_channels:
+        if not searchString:
+            searchString = channel_name
+        elif len(searchString) + len(channel_name) < 600:
```
anderslindho: magic variable

server/recceiver/cfstore.py
```diff
@@ -474,93 +472,114 @@ def create_time_property(owner: str, time: str) -> CFProperty:
...
+    existingChannels = {}
 
-    if ioc_info.hostname is None or ioc_info.iocname is None:
-        raise Exception(f"Missing hostName {ioc_info.hostname} or iocName {ioc_info.iocname}")
+    """
+    The list of pv's is searched keeping in mind the limitations on the URL length
+    The search is split into groups to ensure that the size does not exceed 600 characters
+    """
+    searchStrings = []
+    searchString = ""
```
anderslindho: What's up with the Java style naming all of a sudden?

anderslindho (commented):
I'll end the review here, as it looks unfinished from 2add2302bc99fa8bbbac98c8404272d1137e2505

server/recceiver/cfstore.py
jacomago: Was trying to get the type checker happy with the next line.

server/recceiver/cfstore.py
jacomago: Added to demo.conf and full.conf

jacomago:
For ESS folks, solves CHFI-192 and 198

server/demo.conf
```diff
@@ -66,7 +66,7 @@
 [cf]
 # cf-store application
 # a space-separated list of infotags to set as CF Properties
-#infotags = archive foo bar blah
+#infotags = archive, foo, bar, blah
```
tynanford, line 69: are the commas required?

tynanford:
as the resident ancient python supporter, dataclasses are >= 3.7 and are backported to 3.6 via the pip package. That is fine with me but I thought it is worth mentioning for the 3.6 crowd

server/demo.conf
jacomago: Ah, no. Good point. In fact "x,y z"  would give ["x,y", "z"]

anderslindho:
> as the resident ancient python supporter, dataclasses are >= 3.7 and are backported to 3.6 via the pip package. That is fine with me but I thought it is worth mentioning for the 3.6 crowd

I think this is an interesting topic, @tynanford: should we continue to support EOL python versions for our new releases?

server/recceiver/cfstore.py
```diff
@@ -20,37 +19,244 @@
...
-RECCEIVERID_KEY = "recceiverID"
 RECCEIVERID_DEFAULT = socket.gethostname()
+DEFAULT_MAX_CHANNEL_NAME_QUERY_LENGTH = 600
+
+
+class PVStatus(enum.Enum):
+    """PV Status values."""
+
+    Active = enum.auto()
+    Inactive = enum.auto()
```
anderslindho: Enum states are typically ALL_CAPS

server/recceiver/cfstore.py
```diff
@@ -20,37 +19,244 @@
...
+    environment_variables: str = ""
+    info_tags: str = ""
+    ioc_connection_info: bool = True
+    record_description_enabled: bool = False
+    clean_on_start: bool = True
+    clean_on_stop: bool = True
+    username: str = "cfstore"
+    recceiver_id: str = RECCEIVERID_DEFAULT
+    timezone: Optional[str] = None
+    cf_query_limit: int = 10000
```
anderslindho: Prefer this as global var

server/recceiver/cfstore.py
```diff
@@ -20,37 +19,244 @@
...
+            environment_variables=conf.get("environment_vars", ""),
+            info_tags=conf.get("infotags", ""),
+            ioc_connection_info=conf.get("iocConnectionInfo", True),
+            record_description_enabled=conf.get("recordDesc", False),
+            clean_on_start=conf.get("cleanOnStart", True),
+            clean_on_stop=conf.get("cleanOnStop", True),
+            username=conf.get("username", "cfstore"),
+            recceiver_id=conf.get("recceiverId", RECCEIVERID_DEFAULT),
+            timezone=conf.get("timezone", ""),
+            cf_query_limit=conf.get("findSizeLimit", 10000),
```
anderslindho: Prefer fallback as global var

server/recceiver/cfstore.py
```diff
@@ -20,37 +19,244 @@
...
+class CFProperty:
+    name: str
+    owner: str
+    value: Optional[str] = None
+
+    def as_dict(self) -> Dict[str, str]:
+        """Convert to dictionary for Channelfinder API."""
+        return {"name": self.name, "owner": self.owner, "value": self.value or ""}
+
+    @staticmethod
```
anderslindho: As mentioned before, these type of alternate constructors are typically **class**methods (and then you can also replace `return CFProperty(...)` with `return cls(...)`, but that's optional). A static method is usually just a namespaced method with a loose relationship to the cls.

Same comment applies to all the ones following this

server/recceiver/cfstore.py
```diff
@@ -20,37 +19,244 @@
...
+
+
+@dataclass
+class IocInfo:
+    """Information about an IOC instance."""
+
+    host: str
+    hostname: str
+    ioc_name: str
+    ioc_IP: str
```
anderslindho, line 197: This seems a bit odd. AFAIK recsync with its native definitions does `hostName`, `iocId`. How come you went for `ioc_IP`?

I don't think IOC-name is a concept outside of ESS?

server/recceiver/cfstore.py
```diff
@@ -20,37 +19,244 @@
...
+
+    def as_dict(self) -> Dict[str, Any]:
+        """Convert to dictionary for conversion to json in Channelfinder API."""
+        return {
+            "name": self.name,
+            "owner": self.owner,
+            "properties": [p.as_dict() for p in self.properties],
+        }
+
+    @staticmethod
```
anderslindho: class method

server/recceiver/cfstore.py
```diff
@@ -20,37 +19,244 @@
...
+        self.cf_config = CFConfig.loads(conf)
+        _log.info("CF_INIT %s", self.cf_config)
+        self.name = name
+        self.channel_ioc_ids: Dict[str, List[str]] = defaultdict(list)
+        self.iocs: Dict[str, IocInfo] = dict()
+        self.client: Optional[ChannelFinderClient] = None
+        self.current_time: Callable[[Optional[str]], str] = get_current_time
+        self.lock: DeferredLock = DeferredLock()
 
     def startService(self):
```
anderslindho, line 259: Should we come back to rename remaining camelCase'd methods/functions later on? 💭

server/recceiver/cfstore.py
```diff
@@ -103,244 +310,281 @@ def _startServiceWithLock(self):
...
-        for record_id, (info) in recordInfo.items():
-            if info["pvName"] in recordInfoByName:
-                _log.warning(
-                    "Commit contains multiple records with PV name: {pv} ({iocid})".format(
-                        pv=info["pvName"], iocid=iocid
-                    )
-                )
-                continue
-            recordInfoByName[info["pvName"]] = info
+    def record_info_by_name(self, record_infos: Dict[str, RecordInfo], ioc_info: IocInfo) -> Dict[str, RecordInfo]:
```
anderslindho: I think this maybe should be a static method ?

server/recceiver/cfstore.py
```diff
@@ -103,244 +310,281 @@ def _startServiceWithLock(self):
...
-            iocid,
-            owner,
-            time,
+                # In case, alias exists
+                if self.cf_config.alias_enabled:
+                    if record_name in record_info_by_name:
+                        for record_aliases in record_info_by_name[record_name].aliases:
+                            self.remove_channel(record_aliases, iocid)
+
+    def _commitWithThread(self, transaction: CommitTransaction):
```
anderslindho: Didn't wanna rename also this?

server/recceiver/cfstore.py
```diff
@@ -103,244 +310,281 @@ def _startServiceWithLock(self):
...
+            ioc_name=transaction.client_infos.get("IOCNAME") or str(transaction.source_address.port),
+            ioc_IP=transaction.source_address.host,
+            owner=(
+                transaction.client_infos.get("ENGINEER")
+                or transaction.client_infos.get("CF_USERNAME")
+                or self.cf_config.username
+            ),
+            time=self.current_time(self.cf_config.timezone),
+            port=transaction.source_address.port,
+            channelcount=0,
```
anderslindho: Could/should this be made default?

server/recceiver/cfstore.py
```diff
@@ -103,244 +310,281 @@ def _startServiceWithLock(self):
...
+            recordName: The name of the record to remove.
+            iocid: The IOC ID of the record to remove from.
+        """
+        self.channel_ioc_ids[recordName].remove(iocid)
         if iocid in self.iocs:
-            self.iocs[iocid]["channelcount"] -= 1
-        if self.iocs[iocid]["channelcount"] == 0:
+            self.iocs[iocid].channelcount -= 1
+        if self.iocs[iocid].channelcount == 0:
             self.iocs.pop(iocid, None)
```
anderslindho: Why default `None` after we've already asserted it exists (we would get an exception on the access otherwise)?

server/recceiver/cfstore.py
```diff
@@ -355,460 +599,688 @@ def clean_service(self):
...
+        """
+        return [
+            CFChannel.from_channelfinder_dict(ch)
+            for ch in self.client.findByArgs(
+                prepareFindArgs(
+                    self.cf_config,
+                    [
+                        (CFPropertyName.pvStatus.name, PVStatus.Active.name),
+                        (CFPropertyName.recceiverID.name, recceiverid),
+                    ],
```
anderslindho: kwarg would be nice 🔧

server/recceiver/cfstore.py
```diff
@@ -355,460 +599,688 @@ def clean_service(self):
...
+                            cf_channel,
+                        ),
+                        alias_channel,
+                        processor.managed_properties,
+                    )
+                    channels.append(alias_channel)
+                    _log.debug("Add existing alias %s to previous IOC: %s", alias_channel, last_alias_ioc_id)
+
+
+def orphan_channel(
```
anderslindho, line 715: The naming of all of these functions feels a bit odd.

Also, should they all just be module level methods?

server/recceiver/cfstore.py
```diff
@@ -355,460 +599,688 @@ def clean_service(self):
...
+        for found_channel in client.findByArgs(prepareFindArgs(cf_config, [("~name", each_search_string)])):
+            existing_channels[found_channel["name"]] = CFChannel.from_channelfinder_dict(found_channel)
+        if processor.cancelled:
+            raise defer.CancelledError(
+                f"CF Processor is cancelled, while searching for existing channels: {each_search_string}"
+            )
+    return existing_channels
+
+
+def handle_old_channels(
```
anderslindho: shouldn't this be something like just `handle_channels`, seeing as there's one branch that handles both old and new?

server/recceiver/cfstore.py
anderslindho: Still seems to apply

anderslindho (commented):
Some of the old comments haven't yet been addressed, so please go over those and cross-check/follow-up.

I'd like some of the "function or method or classmethod or staticmethod" stuff to be addressed. I think we need to follow this up with an architectural redesign of modules(+subpackages), because this is already confused and getting worse.

Not sure if we should try to fix all of sonar's complaints now or accept as technical debt to be addressed later.

server/recceiver/cfstore.py
jacomago: This one overrides Twisted.Service.startService so can't be renamed. I'll add a comment.

anderslindho: https://docs.python.org/3/library/typing.html#typing.override - better than a comment IMO

server/recceiver/cfstore.py
jacomago: That's true, but I'm mostly using it to get the names as strings. So I would have to have

```
class PVStatus(enum.StrEnum):
   ACTIVE = "Active"
   INACTIVE = "Inactive"
```

anderslindho: Can also do `.title()`:
```python
>>> import enum
>>> class PVStatus(enum.StrEnum):
...    ACTIVE = enum.auto()
...    INACTIVE = enum.auto()
...
>>> PVStatus.ACTIVE.name.title()
'Active'
```

server/recceiver/cfstore.py
jacomago: *shrug* 
ioc_name has been in the module a long time.

server/recceiver/cfstore.py
jacomago: Probably not. But a lot in this refactor.

server/recceiver/cfstore.py
anderslindho: Thinking more about this, `StrEnum` is def the way to go - don't think we have any other option for `CFPropertyName`.

server/recceiver/cfstore.py
```diff
@@ -20,37 +19,249 @@
 from twisted.internet.threads import deferToThread
 
 from . import interfaces
+from .interfaces import CommitTransaction
+from .processors import ConfigAdapter
```
anderslindho, line 23: We should prob convert the project to use abs imports - could be done together with splitting up `cfstore.py` 💭

server/recceiver/cfstore.py
```diff
@@ -20,37 +19,249 @@
...
+    record_description_enabled: bool = False
+    clean_on_start: bool = True
+    clean_on_stop: bool = True
+    username: str = "cfstore"
+    recceiver_id: str = RECCEIVERID_DEFAULT
+    timezone: Optional[str] = None
+    cf_query_limit: int = DEFAULT_QUERY_LIMIT
+
+    @classmethod
+    def loads(cls, conf: ConfigAdapter) -> "CFConfig":
```
anderslindho, line 59: `loads` usually means "load from **s**tring" (and then `load` is typically from file), but I guess this is OK because `from_configadapter` would be a bit odd/is not immediately clear (that class is a bit strange TBH)

server/recceiver/cfstore.py
```diff
@@ -20,37 +19,249 @@
...
+    name: str
+    owner: str
+    value: Optional[str] = None
+
+    def as_dict(self) -> Dict[str, str]:
+        """Convert to dictionary for Channelfinder API."""
+        return {"name": self.name, "owner": self.owner, "value": self.value or ""}
+
+    @classmethod
+    def from_channelfinder_dict(cls, prop_dict: Dict[str, str]) -> "CFProperty":
```
anderslindho: No need to specify `channelfinder` when the dict keys maps exactly to members here IMO; now you're also breaking the `as_dict` -> `from_dict` symmetry

server/recceiver/cfstore.py
```diff
@@ -20,37 +19,249 @@
...
+class RecordInfo:
+    """Information about a record to be stored in Channelfinder."""
+
+    pv_name: str
+    record_type: Optional[str] = None
+    info_properties: List[CFProperty] = field(default_factory=list)
+    aliases: List[str] = field(default_factory=list)
+
+
+class CFPropertyName(enum.Enum):
```
anderslindho: `StrEnum` 🤷

server/recceiver/cfstore.py
```diff
@@ -20,37 +19,249 @@
...
+    def as_dict(self) -> Dict[str, Any]:
+        """Convert to dictionary for conversion to json in Channelfinder API."""
+        return {
+            "name": self.name,
+            "owner": self.owner,
+            "properties": [p.as_dict() for p in self.properties],
+        }
+
+    @classmethod
+    def from_channelfinder_dict(cls, channel_dict: Dict[str, Any]) -> "CFChannel":
```
anderslindho: Same as before: should just be `from_dict`

server/recceiver/cfstore.py
```diff
@@ -103,244 +315,285 @@ def _startServiceWithLock(self):
...
+                record_property_names_list = [s.strip(", ") for s in self.cf_config.info_tags.split()]
+                if self.cf_config.record_description_enabled:
+                    record_property_names_list.append(CFPropertyName.recordDesc.name)
                 # Are any required properties not already present on CF?
-                properties = required_properties - set(cf_properties)
+                properties = required_properties - cf_properties
                 # Are any whitelisted properties not already present on CF?
                 # If so, add them too.
-                properties.update(set(record_property_names_list) - set(cf_properties))
+                properties.update(set(record_property_names_list) - cf_properties)
```
anderslindho: Could just make `record_property_names_list` a set at creation

server/recceiver/cfstore.py
```diff
@@ -355,460 +608,687 @@ def clean_service(self):
...
 
-    if hostName is None or iocName is None:
-        raise Exception("missing hostName or iocName")
+    if iocid not in iocs:
+        _log.warning("IOC Env Info %s not found in ioc list: %s", ioc_info, iocs)
 
-    if processor.cancelled:
-        raise defer.CancelledError()
+    if ioc_info.hostname is None or ioc_info.ioc_name is None:
+        raise Exception(f"Missing hostName {ioc_info.hostname} or iocName {ioc_info.ioc_name}")
```
anderslindho: Prefer explicit exception. Can maybe use standard lib `AttributeError` or `ValueError` here, if you don't want to create custom one(s).

server/recceiver/cfstore.py
```diff
@@ -355,460 +608,687 @@ def clean_service(self):
...
+        last_ioc_info.ioc_id,
     )
 
 
 def __merge_property_lists(
-    newProperties: List[Dict[str, str]], channel: Dict[str, List[Dict[str, str]]], managed_properties: Set[str] = set()
-) -> List[Dict[str, str]]:
-    """
-    Merges two lists of properties ensuring that there are no 2 properties with
+    newProperties: List[CFProperty], channel: CFChannel, managed_properties: Set[str] = set()
```
anderslindho: Might as well make this PEP compliant while you're it (-> `new_properties`) - I doubt anything calls it by kwarg

server/recceiver/cfstore.py
```diff
@@ -355,460 +608,687 @@ def clean_service(self):
...
+    Args:
+        newProperties: The new properties.
+        channel: The channel.
+        managed_properties: The managed properties
     """
-    newPropNames = [p["name"] for p in newProperties]
-    for oldProperty in channel["properties"]:
-        if oldProperty["name"] not in newPropNames and (oldProperty["name"] not in managed_properties):
+    newPropNames = [p.name for p in newProperties]
+    for oldProperty in channel.properties:
```
anderslindho: Also here

server/recceiver/cfstore.py
```diff
@@ -355,460 +608,687 @@ def clean_service(self):
...
+        timezone: The timezone.
+    """
     if timezone:
         return str(datetime.datetime.now().astimezone())
     return str(datetime.datetime.now())
 
 
-def prepareFindArgs(conf, args, size=0):
-    size_limit = int(conf.get("findSizeLimit", size))
+def prepareFindArgs(cf_config: CFConfig, args, size=0) -> List[Tuple[str, str]]:
```
anderslindho: Also this function name

anderslindho (commented):
It's pretty good, but IMO we should be fixing these remaining idiomatic python/PEP-8 issues before merge. Can all be in a single new commit if you want, TBH.

anderslindho (approved):

jacomago:
@anderslindho I dropped the override commit, only supported on 3.12 and above.

server/recceiver/cfstore.py
simon-ess: I do agree that I expect `loads` to be from a string, but I would otherwise interpret `load` as just kind of _generic_. So maybe it would be better in this case?

server/recceiver/cfstore.py
```diff
@@ -20,37 +19,248 @@
...
+        """Create CFProperty from Channelfinder json output.
+
+        Args:
+            prop_dict: Dictionary representing a property from Channelfinder.
+        """
+        return cls(
+            name=prop_dict.get("name", ""),
+            owner=prop_dict.get("owner", ""),
+            value=prop_dict.get("value"),
+        )
```
simon-ess, line 102: Do you really need this function? You can also allow defaults in the class definition and just do
```py
CFProperty(**some_dict)
```

server/recceiver/cfstore.py
```diff
@@ -20,37 +19,248 @@
...
+        """Create CFChannel from Channelfinder json output.
+
+        Args:
+            channel_dict: Dictionary representing a channel from Channelfinder.
+        """
+        return cls(
+            name=channel_dict.get("name", ""),
+            owner=channel_dict.get("owner", ""),
+            properties=[CFProperty.from_dict(p) for p in channel_dict.get("properties", [])],
+        )
```
simon-ess, line 236: Same question as above.

server/recceiver/cfstore.py
```diff
@@ -61,39 +271,40 @@ def startService(self):
...
-                    "iocid",
-                    "iocIP",
-                    RECCEIVERID_KEY,
+                    CFPropertyName.HOSTNAME.value,
+                    CFPropertyName.IOC_NAME.value,
+                    CFPropertyName.IOC_ID.value,
+                    CFPropertyName.IOC_IP.value,
+                    CFPropertyName.PV_STATUS.value,
+                    CFPropertyName.TIME.value,
+                    CFPropertyName.RECCEIVER_ID.value,
```
simon-ess, line 300: Do we actually need the `.value`s here? Can we not just work with the enum entities?

simon-ess: This is kind of a general question tbh - everywhere I saw the enums used they were used as their `.value` version. Can't they be let be _except_ when you are printing them?

tynanford:
> > as the resident ancient python supporter, dataclasses are >= 3.7 and are backported to 3.6 via the pip package. That is fine with me but I thought it is worth mentioning for the 3.6 crowd
> 
> I think this is an interesting topic, @tynanford: should we continue to support EOL python versions for our new releases?

Personally I think there is value supporting 3.6 for the user community. It is EOL but it is the default python version on redhat and the RHEL variants. Redhat backports security fixes to 3.6 and will until 2029. Not saying we need to continue 3.6 support until then but I also don't see a big reason to drop it now

server/recceiver/cfstore.py
jacomago: Its the frustrating 'channels' field in the prop_dict :( that blocks doing it that way.

server/recceiver/cfstore.py
jacomago: CFPropertyName is a tricky case as you mentioned, in that this is the 'default' property names. You can have them for the env variables and the info tags.

jacomago:
@anderslindho name to value in https://github.com/ChannelFinder/recsync/compare/26d6654a5786ff342ad5928093d30ab2e37cd0ab..f74749eb22ba3c1f833fb382dbd903f3f8c7b865 is the only change

anderslindho (approved):

server/demo.conf
tynanford: if they aren't required then I would remove them to avoid confusion or change the comment above to say space or comma separated

tynanford:
It looks like StrEnum was introduced in python 3.11 so this breaks for all python versions before then

jacomago:
> It looks like StrEnum was introduced in python 3.11 so this breaks for all python versions before then

Looks like we could use https://pypi.org/project/StrEnum/ but it still wouldn't support 3.6...

PR #125, opened 2025-11-10
Move an IOC to another host test

jacomago:


server/tests/docker.py
```diff
@@ -72,3 +73,27 @@ def start_container(
...
+
+def clone_container(
+    compose: DockerCompose,
+    new_host_name: str,
+    host_name: Optional[str] = None,
+    container_id: Optional[str] = None,
+    sleep_time=10,
+) -> str:
+    container_id = container_id or compose.get_container(host_name).ID
+    if container_id:
```
anderslindho: Generally prefer inverting these:
```
if not container_id:
  raise Exception()
# all the other stuff
```
🔧

server/tests/docker.py
```diff
@@ -72,3 +73,27 @@ def start_container(
...
+        docker_client = DockerClient()
+        container = docker_client.containers.get(container_id)
+        image = container.image
+        networks = container.attrs["NetworkSettings"]["Networks"].keys()
+        container.stop()
+        time.sleep(sleep_time)
+        container.remove()
+        docker_client.containers.run(
+            image, detach=True, environment={"IOC_NAME": host_name}, hostname=new_host_name, network=list(networks)[0]
```
anderslindho: Doesn't matter hugely in this context (due to small nbr of entities), but prefer `next(iter(sequence))` over `list(sequence)[0]` 🔧

anderslindho (approved):

PR #126, opened 2025-11-11
Remove things tests

jacomago:
Tests for 

Removing a channel
Removing an alias


anderslindho (commented):
Please reconsider the commit structure or expand the commit messages a bit

client/demoApp/Db/test_remove_infotag.db
```diff
@@ -8,3 +8,9 @@ record(longout, "$(P)lo:base_record") {
   info("hello", "world")
   info("archive", "default")
 }
+
+record(longout, "$(P)lo:base_pv2") {
+  info("test", "testing")
+  info("hello", "world")
+  info("archive", "default")
+}
```
simon-ess: This is a somewhat confusing change on its own because it _really doesn't look like you're removing an infotag_, but the point I guess is that since you have added the PV to the base .db file you need it here as well.

It might be nice if there were a better way to manage this, but *shrug*. Maybe something with `dbLoadTemplate`, or one file per record?

client/.gitignore
```diff
@@ -7,7 +7,6 @@ envPaths
 configure/*.local
 
 dbd/
-db/
```
tynanford, line 10: why is this removed?

client/iocBoot/iocdemo/st_base.cmd
```diff
@@ -0,0 +1,35 @@
...
+addReccasterEnvVars("CONTACT", "SECTOR")
+addReccasterEnvVars("BUILDING")
+
+addReccasterExcludePattern("*_", "*__")
+addReccasterExcludePattern("*exclude_this")
+
+## Load record instances
+dbLoadRecords("../../db/reccaster.db", "P=$(IOCSH_NAME):")
+dbLoadRecords("../../db/somerecords.db","P=$(IOCSH_NAME):")
+dbLoadRecords("../../db/test_base.db", "P=$(IOCSH_NAME):")
```
anderslindho: Hmm, so this seems to be a copy of `st.cmd`, while the associated removal of `st_test.cmd` to me - without additional context - suggested that this was a rename of that file

server/tests/test_bash_ioc.py
```diff
@@ -62,21 +87,10 @@ def get_len_archive_properties(base_channel):
...
-        docker_exec_new_command(docker_ioc, "./demo /recsync/iocBoot/iocdemo/st_remove_infotag.cmd")
-        # Detach by not waiting for the thread to finish
-
-        LOG.debug("ioc1-1 restart")
-        assert wait_for_sync(cf_client, lambda cf_client: check_channel_property(cf_client, name=base_channel_name))
-        LOG.debug("ioc1-1 has restarted and synced")
+        # Act
+        restart_ioc(docker_ioc, cf_client, base_channel_name, "st_remove_infotag")
 
+        # Assert
```
anderslindho, line 153: This commit is a nice change, but I'd say it's a matter of factoring out start and restart functions 🔧

server/tests/client_checks.py
```diff
@@ -12,6 +12,9 @@
 MAX_WAIT_SECONDS = 180
 TIME_PERIOD_INCREMENT = 2
 
+# (2 reccaster.db, 2 somerecords.db, 2 aliases in somerecords.db, 2 in test_base.db)
+BASE_IOC_CHANNEL_COUNT = 2 + (2 + 2) + 2
```
anderslindho: A bit confusing that BASE_COUNT doesn't count what comes from *base.db IMO

server/tests/test_multiple_recceiver.py
```diff
@@ -32,7 +32,7 @@ def test_number_of_channels_and_channel_name(self, cf_client: ChannelFinderClien
     # Smoke Test Default Properties
     def test_number_of_aliases_and_alais_property(self, cf_client: ChannelFinderClient) -> None:
         channels = cf_client.find(property=[("alias", "*")])
-        assert len(channels) == 8
+        assert len(channels) == IOC_COUNT * BASE_ALIAS_COUNT
```
anderslindho, line 35: Can squash this commit with previous

client/demoApp/Db/test_remove_infotag.db
anderslindho: Agree - I think one record per file might help here

Could also template startup scripts, which then further could "automate" getting record counts (as each dbLoadTemplate would be +=1)

But maybe too much indirection - not sure

anderslindho (commented):
I think the names - of startup scripts, db files, records - all are a bit confusing. What is "base"? Why do you sometimes use the term "pv" and sometimes "record"? What even is a base record - are you claming that the collection of "test", "hello", "archive" info-tags with those values are the base of the setup? (Then why is `ai:base_pv` different?)

I also think a refactoring along what @simon-ess suggested would make sense, to have one record per db file. I would take it further and template startup scripts in the test file where they are relevant, and I might even define the db-files in the same manner (using tmp path/dir fixtures).

client/.gitignore
jacomago: I think I need to rethink this, but generally there are db files inside the git repo which should be tracked.

simon-ess: You can still track them even though you have that in gitignore. You just need to do a `git add -f foo`

anderslindho: But why set an ignore on a directory that contains files you typically expect to track? 🙃 

You can also just drop temporary test files outside of this directory; maybe change the `client/.gitignore` to do `*.db` (but keep this removal of `db\`)? 🔧

simon-ess: In general EPICS modules do not track the `$(TOP)` output directories; this is correct IMO. I admit that I am surprised that there is a .db file _in that directory_ that you want to track.

mdavidsaver: You might wish to change these entries to eg.

```
/dbd/
/db/
```

to make clear that they apply only to the directory containing `.gitignore`, and not recusively.  As Simon says, these entries are meant to exclude directories containing build output which should not be versioned.

server/tests/test_bash_ioc.py
```diff
@@ -25,7 +27,7 @@
     encoding="utf-8",
 )
 
-setup_compose = ComposeFixtureFactory(Path("tests") / "docker" / "test-bash-ioc.yml").return_fixture()
+setup_compose = ComposeFixtureFactory(Path("docker") / Path("test-bash-ioc.yml")).return_fixture()
```
tynanford: the CI tests are failing and this is in the logs:

```
stderr = b'open /home/runner/work/recsync/recsync/server/docker/test-bash-ioc.yml: no such file or directory\n'
```

jacomago: Nice catch!

tynanford:
We still have the issue with flaky test failures. I assume the 3.9 and 3.12 tests will pass if we re-run them a couple of times. Doesn't have to be part of this PR but I wonder if we could change the port allocation to be a random port in `cf-compose.yml`

```
ports:
  - "0:8080" 
```

And then if the `get_service_host_and_port` call in `client_checks.py` works like it sounds, it will create `cf_client` with the port docker assigned it?

```
ERROR    testcontainers.compose.compose:compose.py:467 STDERR:
 recc1  Built
 ioc1-1  Built
 Container test-single-ioc-elasticsearch-1  Running
 Container test-single-ioc-elasticsearch-1  Waiting
 Container test-single-ioc-elasticsearch-1  Healthy
 Container test-single-ioc-cf-1  Starting
Error response from daemon: failed to set up container networking: driver failed programming external connectivity on endpoint test-single-ioc-cf-1 (8fa9cb4b3ba67cc61319d41a149350d1860d683fbab0bdd888cdb9cedeceb083): Bind for 0.0.0.0:8080 failed: port is already allocated
```

Or maybe the containers just need to be torn down at the end of each test?

jacomago:
> We still have the issue with flaky test failures. I assume the 3.9 and 3.12 tests will pass if we re-run them a couple of times. Doesn't have to be part of this PR but I wonder if we could change the port allocation to be a random port in `cf-compose.yml`
> 
> ```
> ports:
>   - "0:8080" 
> ```
> 
> And then if the `get_service_host_and_port` call in `client_checks.py` works like it sounds, it will create `cf_client` with the port docker assigned it?
> 
> ```
> ERROR    testcontainers.compose.compose:compose.py:467 STDERR:
>  recc1  Built
>  ioc1-1  Built
>  Container test-single-ioc-elasticsearch-1  Running
>  Container test-single-ioc-elasticsearch-1  Waiting
>  Container test-single-ioc-elasticsearch-1  Healthy
>  Container test-single-ioc-cf-1  Starting
> Error response from daemon: failed to set up container networking: driver failed programming external connectivity on endpoint test-single-ioc-cf-1 (8fa9cb4b3ba67cc61319d41a149350d1860d683fbab0bdd888cdb9cedeceb083): Bind for 0.0.0.0:8080 failed: port is already allocated
> ```
> 
> Or maybe the containers just need to be torn down at the end of each test?

They should be torn down between each test class already. I think this problem is them not tearing down fast enough. I will try random port in another PR.

The previous flakyness I used to see was the docker machine not being ready at the time the tests started, which I'm less sure on how to fix as its from the github actions os.

server/tests/client_checks.py
```diff
@@ -13,6 +13,10 @@
 TIME_PERIOD_INCREMENT = 2
 DEFAULT_CHANNEL_NAME = "IOC1-1:ai:test"
 
+# 1 record and 1 alias
+BASE_ALIAS_COUNT = 1
+BASE_IOC_CHANNEL_COUNT = 2
```
anderslindho: You can skip the comment if you do

```suggestion
BASE_CHANNEL_COUNT = 1
BASE_ALIAS_COUNT = 1
BASE_IOC_CHANNEL_COUNT = BASE_CHANNEL_COUNT + BASE_ALIAS_COUNT
```

🔧

server/tests/test_bash_ioc.py
```diff
@@ -46,43 +48,105 @@ def stream_logs(exec_result, cmd: str):
...
+        lambda cf_client: check_channel_property(cf_client, name=channel_name, prop=INACTIVE_PROPERTY),
+    )
+    ioc_container.start()
+
+    docker_exec_new_command(ioc_container, "./demo /ioc/st.cmd", env={"DB_FILE": new_db_file})
+    # Detach by not waiting for the thread to finish
+
+    LOG.debug("ioc1-1 restart")
+    assert wait_for_sync(cf_client, lambda cf_client: check_channel_property(cf_client, name=channel_name))
+    LOG.debug("ioc1-1 has restarted and synced")
```
anderslindho: You can also use the assert message and invert these;
```python
assert wait_for_sync(cf_client, lambda cf_client: check_channel_property(cf_client, name=channel_name)), \
    "ioc1-1 failed to restart and sync"
```

🔧

server/tests/test_bash_ioc.py
```diff
@@ -46,43 +48,105 @@ def stream_logs(exec_result, cmd: str):
...
+
+
+class TestRemoveInfoTag:
+    def test_remove_infotag(self, setup_compose: DockerCompose) -> None:  # noqa: F811
```
anderslindho: Why need noqa on F811 here? 🤔

server/tests/test_bash_ioc.py
```diff
@@ -46,43 +48,105 @@ def stream_logs(exec_result, cmd: str):
...
         # Check ioc1-1 has ai:test with info tag "archive"
         LOG.debug('Checking ioc1-1 has ai:test with info tag "archive"')
-        channel = cf_client.find(name=DEFAULT_CHANNEL_NAME)
+        channels = cf_client.find(name=DEFAULT_CHANNEL_NAME)
+        TEST_INFO_TAG = {"name": "archive", "owner": "admin", "value": "testing", "channels": []}
 
-        def get_len_archive_properties(channel):
-            return len([prop for prop in channel[0]["properties"] if prop["name"] == "archive"])
+        assert len(channels) == test_channel_count
+        assert TEST_INFO_TAG in channels[0]["properties"]
```
anderslindho: Sequence index access is strange here, as well as on row 104. If you want to assert that there's (/not) `any` `TEST_INFO_TAG` in properties, or if you want to assert that there's exactly 1 of those (or not), you should write out that check

jacomago: Not sure what you mean, you want

any([prop == TEST_INFO_TAG for prop in channels[0]["properties"]) 

?

anderslindho: If you want 1 or more, yes, but with `for c in channels` so we skip doing `channels[0]`

anderslindho (approved):

jacomago:
> We still have the issue with flaky test failures. I assume the 3.9 and 3.12 tests will pass if we re-run them a couple of times. Doesn't have to be part of this PR but I wonder if we could change the port allocation to be a random port in `cf-compose.yml`
> 
> ```
> ports:
>   - "0:8080" 
> ```
> 
> And then if the `get_service_host_and_port` call in `client_checks.py` works like it sounds, it will create `cf_client` with the port docker assigned it?
> 
> ```
> ERROR    testcontainers.compose.compose:compose.py:467 STDERR:
>  recc1  Built
>  ioc1-1  Built
>  Container test-single-ioc-elasticsearch-1  Running
>  Container test-single-ioc-elasticsearch-1  Waiting
>  Container test-single-ioc-elasticsearch-1  Healthy
>  Container test-single-ioc-cf-1  Starting
> Error response from daemon: failed to set up container networking: driver failed programming external connectivity on endpoint test-single-ioc-cf-1 (8fa9cb4b3ba67cc61319d41a149350d1860d683fbab0bdd888cdb9cedeceb083): Bind for 0.0.0.0:8080 failed: port is already allocated
> ```
> 
> Or maybe the containers just need to be torn down at the end of each test?

Think I will try this now....

PR #127, opened 2025-11-13
Refactor docker ioc used in testing recceiver

jacomago:
Pre PR for #126  based on code review on that one.


server/tests/test_bash_ioc.py
```diff
@@ -18,7 +18,7 @@
     encoding="utf-8",
 )
 
-setup_compose = ComposeFixtureFactory(Path("docker") / Path("test-bash-ioc.yml")).return_fixture()
+setup_compose = ComposeFixtureFactory(Path("tests") / Path("docker") / Path("test-bash-ioc.yml")).return_fixture()
```
simon-ess: You don't need all these additional `Path`s:
```py
Path("tests") / "docker" / "test-bash-ioc.yml"
```
(this is one of my favourite synctactic sugars from pathlib...)

server/tests/test_multiple_recceiver.py
```diff
@@ -12,8 +12,8 @@
 
 RECSYNC_RESTART_DELAY = 30
 # Number of channels expected in the default setup
-# 4 iocs, 6 channels per ioc (2 reccaster.db, 2 somerecords.db, 2 aliases in somerecords.db)
-EXPECTED_DEFAULT_CHANNEL_COUNT = 32
+# 4 iocs, 1 channel 1 alias in archive.db
+EXPECTED_DEFAULT_CHANNEL_COUNT = 4 * 2
 
 setup_compose = ComposeFixtureFactory(Path("tests") / Path("docker") / Path("test-multi-recc.yml")).return_fixture()
```
simon-ess: Path path path path path

server/tests/test_bash_ioc.py
```diff
@@ -54,29 +60,28 @@ def test_remove_property(self, setup_compose: DockerCompose) -> None:  # noqa: F
 
         # Check ioc1-1 has ai:archive with info tag "archive"
         LOG.debug('Checking ioc1-1 has ai:archive with info tag "archive"')
-        archive_channel_name = "IOC1-1:ai:archive"
-        archive_channel = cf_client.find(name=archive_channel_name)
+        channel = cf_client.find(name=DEFAULT_CHANNEL_NAME)
 
-        def get_len_archive_properties(archive_channel):
-            return len([prop for prop in archive_channel[0]["properties"] if prop["name"] == "archive"])
```
simon-ess, line 61: Somewhat out of scope, but I don't like the `[0]` here; in particular given that `channel` is not obviously a list.

server/tests/test_multiple_recceiver.py
```diff
@@ -27,16 +27,16 @@ class TestMultipleRecceiver:
     def test_number_of_channels_and_channel_name(self, cf_client: ChannelFinderClient) -> None:
         channels = cf_client.find(name="*")
         assert len(channels) == EXPECTED_DEFAULT_CHANNEL_COUNT
-        assert channels[0]["name"] == "IOC1-1:ai:archive"
+        assert channels[0]["name"] == DEFAULT_CHANNEL_NAME
```
simon-ess, line 30: Is there a guarantee that the records will be returned from CF in a particular order? It seems potentially fragile to test that you expect 8 channels and that the first one has a specific name.

server/tests/ioc/st.cmd
```diff
@@ -6,18 +6,6 @@ var(reccastTimeout, 5.0)
...
-
-epicsEnvSet("CONTACT", "mycontact")
-epicsEnvSet("BUILDING", "mybuilding")
-epicsEnvSet("SECTOR", "mysector")
-
-addReccasterEnvVars("CONTACT", "SECTOR")
-addReccasterEnvVars("BUILDING")
-
-addReccasterExcludePattern("*_", "*__")
-addReccasterExcludePattern("*exclude_this")
```
simon-ess: Should these definitely be removed? These have been added relatively recently, are they being tested?

server/tests/test_bash_ioc.py
jacomago: I will leave that for #124

server/tests/test_multiple_recceiver.py
jacomago: Yes, it is ordered on channel name.

server/tests/ioc/st.cmd
jacomago: I hope they are tested in the RecCaster code rather than RecCeiver. But sure can add a test after this PR for the exclude pattern.

server/tests/docker/test-bash-ioc.yml
```diff
@@ -1,10 +1,10 @@
 name: test-bash-ioc
 include:
-  - cf-compose.yml
+  - ../../docker/cf-compose.yml
```
anderslindho, line 3: Don't love these relative paths. Can't we assume tests will be run from project root and do something like `/server/compose.yml`?

server/tests/docker/test-bash-ioc.yml
```diff
@@ -28,7 +28,17 @@ services:
...
 networks:
   net-1-recc-1:
     driver: bridge
+
+volumes:
+  ioc:
+    driver: local
+    driver_opts:
+      type: none
+      device: ../ioc
```
anderslindho, line 43: Isn't this fragile? We assume compose is run from a specific dir

server/tests/ioc/st_bugtest.cmd
```diff
@@ -0,0 +1,24 @@
+## Register all support components
```
anderslindho: I guess you do this to clearly show the refactor in the following commit, but it does look a bit odd to add in a commit and delete in the following

server/tests/test_multiple_recceiver.py
```diff
@@ -27,23 +27,23 @@ class TestMultipleRecceiver:
     def test_number_of_channels_and_channel_name(self, cf_client: ChannelFinderClient) -> None:
         channels = cf_client.find(name="*")
         assert len(channels) == EXPECTED_DEFAULT_CHANNEL_COUNT
-        assert channels[0]["name"] == "IOC1-1:Msg-I"
+        assert channels[0]["name"] == "IOC1-1:ai:archive"
```
anderslindho: not really part of this PR, but: if ordering is not part of the test, I would prefer something that ignores it, like
```python
assert any("IOC1-1:ai:archive" in c["name"] for c in channels)
```
(or the more precise `c["name"] == EXP_NAME`)

server/tests/test_bash_ioc.py
anderslindho: FWIW I agree with Simon. If you are making these type of assumptions about the elements in a sequence you can improve the flow to get rid of the 0.

server/tests/test_bash_ioc.py
```diff
@@ -54,29 +60,28 @@ def test_remove_property(self, setup_compose: DockerCompose) -> None:  # noqa: F
...
+        def get_len_archive_properties(channel):
+            return len([prop for prop in channel[0]["properties"] if prop["name"] == "archive"])
 
-        assert get_len_archive_properties(archive_channel) == 1
+        assert get_len_archive_properties(channel) == 1
```
anderslindho, line 69: Don't like this either. You can do this inline and just use sum, e.g.
```python
assert sum(p["name"] == "archive" for p in in channel[0]["properties"])
```
or better yet - to get rid of the 0
```python
assert sum(p["name"] == "archive" for c in channels for p in in c["properties"])
```
(but this reads poorly without line breaks so I'd probably just unpack `channels` into a single var before)

server/tests/test_bash_ioc.py
```diff
@@ -54,29 +60,28 @@ def test_remove_property(self, setup_compose: DockerCompose) -> None:  # noqa: F
...
-        assert wait_for_sync(cf_client, lambda cf_client: check_channel_property(cf_client, name=archive_channel_name))
+        assert wait_for_sync(cf_client, lambda cf_client: check_channel_property(cf_client, name=DEFAULT_CHANNEL_NAME))
         LOG.debug("ioc1-1 has restarted and synced")
 
-        archive_channel = cf_client.find(name=archive_channel_name)
-        LOG.debug("archive channel: %s", archive_channel)
-        assert get_len_archive_properties(archive_channel) == 0
+        channel = cf_client.find(name=DEFAULT_CHANNEL_NAME)
+        LOG.debug("archive channel: %s", channel)
+        assert get_len_archive_properties(channel) == 0
```
anderslindho, line 88: Prefer an (invert check) `all` or (don't) `any`

server/tests/test_multiple_recceiver.py
anderslindho: So if the record is renamed this might break 🙃

anderslindho (commented):
It's an improvement overall, but I would like to see also `server/tests/ioc/*.db` go down from 2 to 1, and it would be nice if we could get rid of the manual `CHANNEL_COUNT`s and instead base that on some param fed to the test setup fn(s)

It's also a bit odd that we have to manually match `CHANNEL_NAME` in tests - you may as well set the entire record name as a macro and just use the same definition

server/tests/docker/test-bash-ioc.yml
jacomago: Leaving as I don't care as long as running pytest works.

server/tests/docker/test-bash-ioc.yml
jacomago: Yes, again. As long as pytest works I don't care.

server/tests/test_multiple_recceiver.py
jacomago: out of scope

server/tests/test_bash_ioc.py
jacomago: out of scope

server/tests/test_bash_ioc.py
jacomago: out of scope

server/tests/ioc/st_bugtest.cmd
jacomago: Yeah, focused on clear commit diffs.

anderslindho (approved):
It's an improvement, but I think my previous summary review comment still stands

PR #129, opened 2025-12-03
remove docker image prune

jacomago:
Removing the docker image prune, as I don't think necessary. Should speed up recceiver tests.

### DIFF
--- a/.github/workflows/server.yml
+++ b/.github/workflows/server.yml
@@ -64,9 +64,6 @@ jobs:
         python -m pip install --upgrade pip
         python -m pip install '.[test]'
         python -m pip install .
-    - name: Clear existing docker image cache
-      shell: bash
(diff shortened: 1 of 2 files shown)


simon-ess:
Two questions:
* How much of a speedup are we expecting?
* Why was this there in the first place? I can see it was added here https://github.com/ChannelFinder/recsync/pull/102 with no comment...

jacomago:
> Two questions:
> 
>     * How much of a speedup are we expecting?
> 

Maybe not much...
>     * Why was this there in the first place? I can see it was added here [End to end testing for RecCeiver. #102](https://github.com/ChannelFinder/recsync/pull/102) with no comment...

I think before building the images would fill the allowed space on the runner.

simon-ess (approved):

PR #130, opened 2025-12-03
use epics-containers base image in client Dockerfile

gilesknap:
as discussed here https://matrix.to/#/!zROwattWUrXjRnikQx:epics-controls.org/$T2sdfO8Y6x3VMkOpWgJnID3n9g5yPp5F4MqyBRTEOqY?via=epics-controls.org&via=matrix.org&via=kit.edu

### DIFF
--- a/client/ioc-compose.yml
+++ b/client/ioc-compose.yml
@@ -4,9 +4,9 @@ services:
     environment:
       - IOCSH_NAME=IOC1
     tty: true
+    stdin_open: true
     networks:
       - net-1-recc-1
-
 networks:
   net-1-recc-1:
     driver: bridge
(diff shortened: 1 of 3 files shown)


gilesknap:
SonarKube's suggestions for the Dockerfile are sensible.

client/ioc-compose.yml
```diff
@@ -1,13 +1,12 @@
-
 services:
   ioc1:
-    build: ../client
+    build: .
```
jacomago, line 3: I'll need to check if this breaks https://github.com/ChannelFinder/recsync/blob/master/server/tests/docker/test-bash-ioc.yml

gilesknap: I did not really need this change, it just looked wrong.
 
(I'm aware that relative paths in nested compose files have odd behaviour)

PR #135, opened 2026-02-06
Set default python version 3.9 and fix strenum problem

jacomago:


jacomago:
I tried to get this check in the pre-commit, but adding all ruff lints wasn't enough, I had to add type checking... Which made the PR huge. So I'll stick with just this change for now.

.pre-commit-config.yaml
```diff
@@ -6,13 +6,18 @@ repos:
       - id: trailing-whitespace
   - repo: https://github.com/astral-sh/ruff-pre-commit
   # Ruff version.
-    rev: v0.9.0
+    rev: v0.15.0
     hooks:
       # Run the linter.
       - id: ruff
         args: [ --fix ]
```
anderslindho, line 13: Don't think this is needed anymore, and the hook has been renamed (ruff-check IIRC)

server/pyproject.toml
```diff
@@ -1,45 +1,54 @@
 [build-system]
-requires = ["setuptools"]
 build-backend = "setuptools.build_meta"
+requires = [ "setuptools" ]
 
 [project]
-name="recceiver"
+name = "recceiver"
+version = "1.5"
+description = "resync server"
```
anderslindho: Typo

server/pyproject.toml
```diff
@@ -1,45 +1,54 @@
...
+  "twisted",
 ]
 
-[project.optional-dependencies]
-test = ["pytest", "testcontainers>=4"]
-
-[project.urls]
-Repository="https://github.com/mdavidsaver/recsync"
+optional-dependencies.test = [ "pytest", "testcontainers>=4" ]
+urls.Repository = "https://github.com/mdavidsaver/recsync"
```
anderslindho: I think this is wrong?

server/pyproject.toml
```diff
@@ -1,45 +1,54 @@
...
-twisted = ["plugins/recceiver_plugin.py"]
-
-[tool.pytest.ini_options]
-log_cli = true
-log_cli_level = "DEBUG"
+twisted = [ "plugins/recceiver_plugin.py" ]
 
 [tool.ruff]
-line-length = 120
+target-version = "py39"
```
anderslindho, line 44: This doesn't mesh well with your attempt to externalise python version into a variable

server/pyproject.toml
```diff
@@ -1,45 +1,54 @@
...
+line-length = 120
 # Enable Pyflakes (`F`) and a subset of the pycodestyle (`E`)  codes by default.
 # Unlike Flake8, Ruff doesn't enable pycodestyle warnings (`W`) or
 # McCabe complexity (`C901`) by default.
-select = ["E", "F", "I"]
+lint.select = [ "E", "F", "I" ]
+
+[tool.pytest.ini_options]
+log_cli = true
+log_cli_level = "DEBUG"
```
anderslindho: Do we really need this as default?

server/pyproject.toml
simon-ess: We can blame that on me!

See https://github.com/ChannelFinder/recsync/blame/master/server/pyproject.toml and previously (10 yrs ago!) https://github.com/ChannelFinder/recsync/blame/88665ce9943d9883f00326fda58c027ad9a050df/server/setup.py

server/pyproject.toml
jacomago: Its used to make the logs published by docker appear in the ci logs.

server/pyproject.toml
jacomago: Not sure I can fix that.

server/pyproject.toml
jacomago: What typo?

simon-ess: "resync" vs "recsync"

.pre-commit-config.yaml
anderslindho: To be clear, my comment was on row 13, that I don't think `args: [--fix]` is needed anymore

server/pyproject.toml
anderslindho: You can abandon the idea to externalise it :)

server/pyproject.toml
anderslindho: Isn't this setting global, i.e. would apply also if I clone and run locally?
Can't it be specified on the command-line when executing pytest?

.pre-commit-config.yaml
jacomago: Still in the docs: https://github.com/astral-sh/ruff-pre-commit

server/pyproject.toml
jacomago: You mean out of the dockerfile?

server/pyproject.toml
jacomago: Yeah, you can override it. But sure tests running with debug enabled makes sense?

server/pyproject.toml
anderslindho: Yeah

server/pyproject.toml
anderslindho: IMO no, since pytest is configured by default to display log messages over a certain verbosity only for failing tests, and since it is an easy modification for any user to alter the verbosity if necessary.

simon-ess: I'm with @jacomago on this one. If the tests are failing for some reason in CI but are not reproducible locally, this is pretty useful.

anderslindho: My point is not to remove it from CI, but that this is a global config. We can still lower the verbosity for CI.

anderslindho: Just change https://github.com/ChannelFinder/recsync/blob/master/.github/workflows/server.yml#L69 to `pytest --log-cli-level=DEBUG`

.github/workflows/server.yml
tynanford, line 21: should remove 3.6 - 3.8 in the CI

server/recceiver/application.py
```diff
@@ -3,13 +3,13 @@
 import logging
 import random
 
-from zope.interface import implementer
```
tynanford, line 6: is moving these imports necessary? same goes for all the other import changes

server/pyproject.toml
```diff
@@ -1,45 +1,50 @@
...
+description = "The recsync server for receiving updates from the 'client' and pushing them to ChannelFinder, or a dbstore."
+readme = "README.md"
 authors = [
-    {name = "Michael Davidsaver", email = "mdavidsaver@gmail.com"},
+  { name = "Michael Davidsaver", email = "mdavidsaver@gmail.com" },
 ]
-description="resync server"
-version="1.5"
-readme = "README.md"
 requires-python = ">=3.6"
```
tynanford, line 13: update to >=3.9

server/pyproject.toml
```diff
@@ -1,45 +1,50 @@
...
+classifiers = [
```
tynanford, line 14: remove 3.6. - 3.8

server/pyproject.toml
```diff
@@ -1,45 +1,50 @@
...
+  "Programming Language :: Python :: 3.7",
+  "Programming Language :: Python :: 3.8",
+  "Programming Language :: Python :: 3.9",
+  "Programming Language :: Python :: 3.10",
+  "Programming Language :: Python :: 3.11",
+  "Programming Language :: Python :: 3.12",
+  "Programming Language :: Python :: 3.13",
+  "Programming Language :: Python :: 3.14",
+]
 dependencies = [
```
tynanford, line 26: with >= 3.9 , we could use StrEnum and use this pip package like you mentioned? https://pypi.org/project/StrEnum/

tynanford (commented):
It's too bad the 1.8 release already included the update that was >= 3.11 but oh well. Let's make a new release after this PR is merged

Also can you update the README with the new minimum python version?

server/pyproject.toml
jacomago: Do you not need 3.6 anymore? I'm happy to remove.

server/recceiver/application.py
jacomago: Its from the latest version of ruff, the formatter must have changed to order the imports differently.

server/pyproject.toml
jacomago: I think using the mixin is fine.

server/pyproject.toml
```diff
@@ -1,45 +1,47 @@
 [build-system]
-requires = ["setuptools"]
 build-backend = "setuptools.build_meta"
+requires = [ "setuptools" ]
 
 [project]
-name="recceiver"
+name = "recceiver"
+version = "1.8-dev"
```
anderslindho: Should we keep this as dev version?

server/pyproject.toml
```diff
@@ -1,45 +1,47 @@
...
+description = "The recsync server for receiving updates from the 'client' and pushing them to ChannelFinder, or a dbstore."
```
anderslindho: Meh description... I don't quite like single-quote `'client'`, and the mention of an undefined-and-difficult-to-interpret `dbstore` is also a bit off. We may as well use the common terms `recCeiver`, `recCaster`, or at least make it clear that _any_ client (considering the python-rust PVA server lib) can forward. Maybe something like:
```suggestion
description = "recCeiver is a server component of the recsync protocol. It receives record updates from recsync clients (e.g., recCasters) and forwards them to a configurable backend such as ChannelFinder."
```

server/pyproject.toml
```diff
@@ -1,45 +1,47 @@
...
+  "twisted",
 ]
 
-[project.optional-dependencies]
-test = ["pytest", "testcontainers>=4"]
-
-[project.urls]
-Repository="https://github.com/mdavidsaver/recsync"
+optional-dependencies.test = [ "pytest", "testcontainers>=4" ]
+urls.Repository = "https://github.com/ChannelFindder/recsync"
```
anderslindho: Typo

server/pyproject.toml
simon-ess: I like this rephrase, although I might prefer "recCeiver is _the_ server component...".

simon-ess (approved):

server/pyproject.toml
jacomago: Well its a question of how to handle a hardcoded version number not linked to the git repo. I figure this is a simple way, then I can make another pr after this one to change to 1.8.1

server/pyproject.toml
jacomago: For now...

anderslindho: My (implicit) point :)

simon-ess: I gathered that from you writing it, but I actually still think it's more correct to say "the". This isn't really due to there being a unique implementation of the server, so much as it is the server in the client/server relationship.

server/pyproject.toml
tynanford: Ah, I read "default" in the PR description as planning to get rid of 3.6-3.8

It's up to the community what to do. I think there's value in making life easiest on the people using/deploying software in these open source projects. So it is good to keep support for older versions when it's not too much hassle and I would say keep 3.6. But I am fine moving to 3.9 too (RHEL 9 default) if no one else wants 3.6 support.

jacomago: I'm wary of going beyond the python supported versions  https://devguide.python.org/versions/. 3.9 is already end of life. Wouldn't be surprised if that starts breaking the ci soon.

tynanford: Eh, I think the hassle is worth it to help the lab setups where there might just be one person trying to maintain all this EPICS software. But just my opinion

ralphlange: +1
At ITER - at 1 million records with less than 10 of ~160 plant systems delivered - we are struggling to migrate everything _to_ RHEL 8 (that uses Python 3.6 as standard).

anderslindho: Side-point, but you might want to have a look at conda(/mamba, etc.) @ralphlange 

Re py3.6, I am a bit torn: there may be sites that are limited to it, but EOL'd more than 4 years ago, and is becoming increasingly costly to maintain support for; that is at least what we have seen for the projects where we have tried to do so.

ralphlange: I know there are better solutions - it's simply not my decision to make.
Almost all ITER plant systems are in-kind contributions, and many take 5-10 years between prototyping and delivery. Which means we're getting deliveries based on versions/technologies of 5-10 years ago.

If supporting old versions is getting costly, drop it. We'll find ways.
Please don't drop it just because it's old.

anderslindho: We discussed it today and said that we either will build (but not test) from 3.6 or else maybe set up a dedicated (limited support) release branch for 3.6.

jacomago:
I think this PR is fine now. The tests look like to do with the ioc container in the tests taking too long to build for some reason. So I will think about a separate PR for that. 

And I locally ran the tests and they passed.

server/tests/client_checks.py
```diff
@@ -67,7 +67,7 @@ def create_client_and_wait(compose: DockerCompose, expected_channel_count) -> Ch
 
 
 def create_client_from_compose(compose: DockerCompose) -> ChannelFinderClient:
-    cf_host, cf_port = compose.get_service_host_and_port("cf")
```
anderslindho, line 70: Magic value - should be moved to variable

anderslindho (approved):

tynanford (approved):

PR #137, opened 2026-03-17
Allow loading config vars from env

jacomago:
This allows overriding the configuration files by reading from the environment any config vars prefixed with RECCEIVER_


server/recceiver/processors.py
```diff
@@ -21,6 +22,11 @@
 ]
 
 
+def _env_vars():
+    prefix = "RECCEIVER_"
+    return {k[len(prefix) :].lower(): v for k, v in os.environ.items() if k.startswith(prefix)}
```
anderslindho: https://docs.python.org/3/library/stdtypes.html#str.removeprefix

server/recceiver/processors.py
```diff
@@ -33,21 +39,21 @@ def __contains__(self, key):
...
         except (ConfigParser.NoOptionError, ValueError):
             return D
 
     def __getitem__(self, key):
-        try:
-            return self._C.get(self._S, key)
-        except ConfigParser.NoOptionError:
+        result = self.get(key)
+        if result is None:
             raise KeyError("No option value")
```
anderslindho, line 56: Can just do `return self[key]` (or is that ValueError?)

server/compose.yml
```diff
@@ -12,11 +12,16 @@ services:
...
       timeout: 30s
       retries: 3
+    environment:
+      RECCEIVER_PROCS: cf
+      RECCEIVER_LOGLEVEL: DEBUG
+      RECCEIVER_INFOTAGS: archive
+      RECCEIVER_ALIAS: "True"
+      RECCEIVER_RECORDTYPE: "True"
+      RECCEIVER_RECORDDESC: "True"
+      RECCEIVER_CLEANONSTART: "True"
```
anderslindho: Hmm, does this mean `"\"True\""`?

It's a bit odd to have env var options as these cases - mixed between lowercase, all caps, camel.

anderslindho (commented):
Isn't the ordering a bit strange here, that env vars - which are implicitly read - are prioritised over a config file?

server/recceiver/processors.py
jacomago: I think it is ConfigParser.NoOptionError

server/recceiver/processors.py
simon-ess: It seems odd to me that we re-fetch the environment vars every time we call e.g. ConfigAdapter.get(), .getBoolean().

jacomago:
> Isn't the ordering a bit strange here, that env vars - which are implicitly read - are prioritised over a config file?

Its the same hierachy as spring-boot https://docs.spring.io/spring-boot/reference/features/external-config.html

server/compose.yml
jacomago: swapped to true

server/recceiver/processors.py
anderslindho: If you did `dict[key]` access with try-except you can get exception chaining, but not sure if that's better or not. Your call.

server/compose.yml
anderslindho: Can we instead do `TRUE` (or `YES` or `1` or whatever) so it looks more like env var values?

server/recceiver/processors.py
jacomago: I think self is not a dictionary. I think it is a wrapper around ConfigParser. I think you have to use get.

server/compose.yml
jacomago: done

anderslindho (approved):

PR #138, opened 2026-03-20
Add an env_owner_variable

jacomago:
This allows to use specific env variable name for the owner in channelfinder.


anderslindho (commented):
Shouldn't the default be ENGINEER to make this change a no-op (then we override for our site)?

anderslindho (approved):

PR #141, opened 2026-03-26
Pass section to env vars

jacomago:
Otherwise you can't actually pass env vars as config


anderslindho (approved):

jacomago:
Double checked the tests locally, looks like its when running multiple is the problem.

PR #143, opened 2026-04-16
fix: extended UDP broadcast to reach physical devices outside the virtual Docker network

ChemistryTobias:
# Issue
I installed the RecCeiver service as follows:

```bash
mkdir -p $HOME/EPICS/docker/RecCeiver
cd $HOME/EPICS/docker/RecCeiver
git clone https://github.com/ChannelFinder/recsync.git
cd recsync/server
docker compose up -d
```

But with the current Docker `compose.yml` implementation, the UDP broadcast is **confined to the virtual Docker network** and only find IOCs within it. 

# Fix
If this is not the intended standard behavior of your Docker compose,  I want to offer a few minor changes, I used in our setup. 

Otherwise, just keep up with your project, really like it :v:

### DIFF
--- a/server/compose.yml
+++ b/server/compose.yml
@@ -8,7 +8,7 @@ services:
         condition: service_healthy
         restart: true
     healthcheck:
-      test: netstat | grep cf
+      test: curl -s -f http://localhost:8080/ChannelFinder
       interval: 10s
       timeout: 30s
       retries: 3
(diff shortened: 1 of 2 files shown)


server/compose.yml
```diff
@@ -8,7 +8,7 @@ services:
         condition: service_healthy
         restart: true
     healthcheck:
-      test: netstat | grep cf
+      test: curl -s -f http://localhost:8080/ChannelFinder
```
jacomago, line 11: This tests channelfinder is up, not recceiver.

server/compose.yml
```diff
@@ -23,16 +23,14 @@ services:
       RECCEIVER_CF_USERNAME: admin
       RECCEIVER_CF_CFUSERNAME: admin
       RECCEIVER_CF_CFPASSWORD: password
-      RECCEIVER_CF_BASEURL: http://cf:8080/ChannelFinder
+      RECCEIVER_CF_BASEURL: http://localhost:8080/ChannelFinder
```
jacomago, line 26: This breaks the integration tests which have "cf" as the hostname for channelfinder.

server/compose.yml
```diff
@@ -23,16 +23,14 @@ services:
...
       RECCEIVER_CF_VERIFYSSL: FALSE
     volumes:
       - type: bind
         source: docker/config/cf.conf
         target: /home/recceiver/channelfinderapi.conf
         read_only: true
-    networks:
-      - net-1-recc-1
-      - net-2-cf
+    network_mode: host
```
jacomago, line 34: I'm not sure about this. This will.mean container IOC's won't have access to recceiver here? 

As mentioned before this will break the integration tests.

PR #144, opened 2026-04-30
refactor(server): extract framework-neutral protocol package

anderslindho:
Lift RecSync wire encoding and decoding out of the Twisted adapter (`recast.py`, `announce.py`) into a new `recceiver.protocol` subpackage so the parsing logic can be reused if we move from Twisted to asyncio (or any other transport) in the future. The Twisted adapter now operates on typed dataclasses with named fields rather than struct slices and integer offsets.

This is preparation work - no functional change except for the standalone bugfix commit.

### DIFF
--- a/server/recceiver/protocol/messages.py
+++ b/server/recceiver/protocol/messages.py
@@ -1,0 +1,287 @@
+# -*- coding: utf-8 -*-
+"""Framework-neutral RecSync TCP protocol messages."""
+
+import struct
+from dataclasses import astuple, dataclass
+from enum import IntEnum
+from typing import ClassVar
+
+PROTO_ID = 0x5243  # b'RC'
+
+
+class RecordKind(IntEnum):
+    RECORD = 0
+    ALIAS = 1
+
+
+class ProtocolError(ValueError):
+    """Raised when a RecSync message violates the wire protocol."""
+
+
+def _require_length(data, expected_length, message_name, exact=True):
+    if exact and len(data) == expected_length:
+        return
+    if not exact and len(data) >= expected_length:
+        return
+
+    relation = "must be" if exact else "must be at least"
+    raise ProtocolError(f"{message_name} {relation} {expected_length} bytes, got {len(data)}")
+
+
+@dataclass(frozen=True)
+class Header:
+    """TCP frame header.
+
+    Wire layout: PROTO_ID(2) + MSG_ID(2) + BLEN(4).
+    """
+
+    payload: ClassVar = struct.Struct("!HHI")
+
+    msg_id: int
+    body_length: int
+
+    def encode(self):
+        return self.payload.pack(PROTO_ID, int(self.msg_id), self.body_length)
+
+    @classmethod
+    def decode(cls, wire_bytes):
+        _require_length(wire_bytes, cls.payload.size, cls.__name__)
+        proto_id, msg_id, body_length = cls.payload.unpack(wire_bytes)
+        if proto_id != PROTO_ID:
+            raise ProtocolError(f"bad protocol id {proto_id:#06x}")
+        return cls(msg_id, body_length)
+
+
+assert Header.payload.size == 8
+
+
+class TCPMessage:
+    """Base class for TCP payloads carried by a RecSync frame."""
+
+    payload = struct.Struct("!")
+    msg_id = None
+
+    def encode(self):
+        return self.payload.pack(*astuple(self))
+
+    def frame(self):
+        body = self.encode()
+        return Header(self.msg_id, len(body)).encode() + body
+
+    @classmethod
+    def decode(cls, body):
+        _require_length(body, cls.payload.size, cls.__name__)
+        return cls(*cls.payload.unpack(body))
+
+
+assert TCPMessage.payload.size == 0
+
+
+class ClientMessage(TCPMessage):
+    """Message sent from RecCaster to RecCeiver (msg_id 0x0XXX)."""
+
+
+class ServerMessage(TCPMessage):
+    """Message sent from RecCeiver to RecCaster (msg_id 0x8XXX)."""
+
+
+@dataclass(frozen=True)
+class ClientGreeting(ClientMessage):
+    """Client greeting sent on connection establishment.
+
+    Wire layout: VERSION(1) + CLIENT_TYPE(1) + pad(2) + SERV_KEY(4).
+    """
+
+    payload: ClassVar = struct.Struct("!BBxxI")
+    msg_id: ClassVar = 0x0001
+
+    version: int
+    client_type: int
+    server_key: int
+
+
+assert ClientGreeting.payload.size == 8
+
+
+@dataclass(frozen=True)
+class Pong(ClientMessage):
+    """Ping response echoing the server's nonce.
+
+    Wire layout: NONCE(4).
+    """
+
+    payload: ClassVar = struct.Struct("!I")
+    msg_id: ClassVar = 0x0002
+
+    nonce: int
+
+
+assert Pong.payload.size == 4
+
+
+@dataclass(frozen=True)
+class AddRecord(ClientMessage):
+    """Registers a new record or alias.
+
+    Wire layout: RECID(4) + KIND(1) + RTLEN(1) + RNLEN(2) [+ RTYPE(RTLEN) + RNAME(RNLEN)].
+    """
+
+    payload: ClassVar = struct.Struct("!IBBH")
+    msg_id: ClassVar = 0x0003
+
+    record_id: int
+    kind: RecordKind
+    record_type: str
+    record_name: str
+
+    @property
+    def is_alias(self):
+        return self.kind == RecordKind.ALIAS
+
+    def encode(self):
+        kind = _record_kind(self.kind)
+        record_type = self.record_type.encode()
+        record_name = self.record_name.encode()
+        _validate_record_lengths(kind, len(record_type), len(record_name))
+        return (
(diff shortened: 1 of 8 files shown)


server/recceiver/protocol/messages.py
```diff
@@ -0,0 +1,252 @@
...
+
+    nonce: int
+
+
+assert Pong.payload.size == 4
+
+
+@dataclass(frozen=True)
+class AddRecord(ClientMessage):
+    # RECID(4) + KIND(1) + RTLEN(1) + RNLEN(2) [+ RTYPE(RTLEN) + RNAME(RNLEN)]
```
jacomago: Maybe these comments could be doc strings.

server/recceiver/recast.py
```diff
@@ -99,43 +74,57 @@ def writePing(self):
...
-        if ctype != 0:
-            _log.error("I don't understand you! {s}".format(s=ctype))
+        try:
+            greeting = messages.ClientGreeting.decode(body)
+        except messages.ProtocolError as exc:
+            _log.error(f"Protocol error! {exc}")
+            self.transport.loseConnection()
+            return
+        if greeting.client_type != 0:
+            _log.error(f"I don't understand you! {greeting.client_type}")
```
jacomago: Maybe a clearer message here.

jacomago (approved):

server/recceiver/recast.py
```diff
@@ -179,7 +179,7 @@ def recvAddRec(self, body):
 
     # 0x0004
     def recvDelRec(self, body):
-        record_id = _ping.unpack(body[: _ping.size])
+        (record_id,) = _ping.unpack(body[: _ping.size])
```
simon-ess: Good thing that:
* tuples are hashable!
* the del record command is literally never used

server/recceiver/protocol/messages.py
```diff
@@ -0,0 +1,252 @@
...
+    def encode(self):
+        return self.payload.pack(PROTO_ID, int(self.msg_id), self.body_length)
+
+    @classmethod
+    def decode(cls, wire_bytes):
+        _require_length(wire_bytes, cls.payload.size, cls.__name__)
+        proto_id, msg_id, body_length = cls.payload.unpack(wire_bytes)
+        if proto_id != PROTO_ID:
+            raise ProtocolError(f"bad protocol id {proto_id:#06x}")
+        return cls(msg_id, body_length)
```
simon-ess, line 52: Do we actually need this class?

Ok, I see how this fits in. But I think that this is just making it all twisty-like again with a weird heirarchy of dependencies that is not easy to follow. Would it make more sense to have this be a subclass of the TCP message?

server/recceiver/protocol/messages.py
```diff
@@ -0,0 +1,252 @@
...
+
+class TCPMessage:
+    """Base class for TCP payloads carried by a RecSync frame."""
+
+    payload = struct.Struct("!")
+    msg_id = None
+
+    def encode(self):
+        return self.payload.pack(*astuple(self))
+
```
simon-ess, line 66: Does this depend on the ordering of the fields in each subclass? Seems a bit dodgy...

server/recceiver/protocol/messages.py
```diff
@@ -0,0 +1,252 @@
...
+        proto_id, msg_id, body_length = cls.payload.unpack(wire_bytes)
+        if proto_id != PROTO_ID:
+            raise ProtocolError(f"bad protocol id {proto_id:#06x}")
+        return cls(msg_id, body_length)
+
+
+assert Header.payload.size == 8
+
+
+class TCPMessage:
```
simon-ess, line 58: Should this also be a dataclass? astuple doesn't apply otherwise.

Also, is this the correct name? It is a TCPMessage, but it seems that this is a bit too generic a name for our purposes.

server/recceiver/protocol/messages.py
```diff
@@ -0,0 +1,252 @@
...
+    """Base class for TCP payloads carried by a RecSync frame."""
+
+    payload = struct.Struct("!")
```
simon-ess, line 61: Should payload be None as well here? We aren't initialising this class on its own ever, do we?

server/recceiver/protocol/messages.py
```diff
@@ -0,0 +1,252 @@
...
+        body = self.encode()
+        return Header(self.msg_id, len(body)).encode() + body
+
+    @classmethod
+    def decode(cls, body):
+        _require_length(body, cls.payload.size, cls.__name__)
+        return cls(*cls.payload.unpack(body))
+
+
+assert TCPMessage.payload.size == 0
```
simon-ess, line 77: I'm a bit surprised at this. Do we want these bare asserts at import time? Isn't it better to have tests for this?

server/recceiver/recast.py
```diff
@@ -144,47 +133,48 @@ def recvPong(self, body):
...
     def recvDelRec(self, body):
-        record_id = _ping.unpack(body[: _ping.size])
-        self.sess.delRecord(record_id)
+        try:
+            record = messages.DelRecord.decode(body)
+        except messages.ProtocolError:
+            _log.error("Ignoring delete record update")
+            return self.getInitialState()
+        self.sess.delRecord(record.record_id)
         return self.getInitialState()
```
simon-ess, line 169: Should we get rid of this message? It isn't used at all and it feels like it's a bit dangerous since it requires knowing the record ID which is kind of not entirely deterministic. 🤡

simon-ess (commented):
Some bits of this seem a bit twisty IMO but I think this is generally an improvement.

server/recceiver/protocol/messages.py
anderslindho: We need it, yes. It is used 2x via `TCPMessage.frame()` in `recast.py`.

And no I don't think it makes sense to have it be a subclass of the TCP message.

server/recceiver/protocol/messages.py
anderslindho: Valid concern in principle but IMO not in practice. `astuple` on a frozen dataclass yields fields in declaration order, and the struct format string is defined right there in the same class body. They must match, and they always do. And then we have the size assert on top of this. All pretty common practices.

server/recceiver/protocol/messages.py
anderslindho: Every subclass is a dataclass. We could turn this into an abstract base class, but I honestly don't see the value added. And I would say the name is appropriate given it's location.

server/recceiver/protocol/messages.py
anderslindho: Feels like we would have to be careful with more things for no real gain if we did - e.g. `payload.size()` would break with `None`.

server/recceiver/protocol/messages.py
anderslindho: This is pretty common. These are not like pytest asserts - this is more like C's static_assert.

server/recceiver/recast.py
anderslindho: That feels out of scope for this MR.

server/recceiver/protocol/messages.py
simon-ess: I think in this case it would be good to block this class from being instantiated, no?

server/recceiver/protocol/messages.py
simon-ess: ... but is it needed? Isn't this really just providing weird additional tests for the behavior of `struct.Struct`?

anderslindho: It's best practice, so I want to keep it

server/recceiver/protocol/messages.py
anderslindho: I am torn. The `astuple` bit actually blocks direct instantiation. And if we did do ABC we should also prob do abstractmethod, @override, etc. - both on this base base class but also on clientmessage and servermessage. And there are solid arguments from Python devs about ABCs, that they are not designed and intended for these cases (even if this maybe has evolved since); technically it is insufficient to just do ABC. But it _does_ signal some sort of intent. The question, however, is whether that intent via decorator actually helps. From my POV, the code itself is clear, and the docstring already makes the hierarchy explicit.

server/recceiver/protocol/messages.py
simon-ess: Where is that seen? Certainly the struct.Struct docs do not mention this. Searching for best practices doesn't yield anything either.

anderslindho: https://github.com/ChannelFinder/recsync/blob/master/server/recceiver/recast.py#L20-L36

https://github.com/ChannelFinder/recsync/pull/51/changes#diff-91af850eef2e8870e18c57a8fcca90f1c4a7e4642dbc5c5f9e4c49eceb5b4e4fR44

I am pretty sure you'd see it if you open up ssl or whatever. This is - as I understand it - basically what `assert` is for.

simon-ess: Ok, fair point in that it already existed in this repo (although I was more curious about where you found best practices for using `struct.Struct`). But it still seems odd to me in that this is really an assertion test of struct.Struct and doesn't IMO add anything here, although I can accept this.

anderslindho: I have both seen and used this myself when I have (for hobby projects) done graphics programming. I have also seen it used in at minimum C/C++ but IIRC also python as a generally recommended defensive programming practice. Akin to null-checking everything.

jacomago (approved):

PR #145, opened 2026-05-04
Add polling configuration options for ChannelFinder updates

jacomago:
- Introduced `pollingRetryCount` and `pollingInfiniteRetries` settings in demo and full configuration files.
- Updated CFConfig class to handle new polling parameters.
- Enhanced polling logic to respect retry count and infinite retry options.

### DIFF
--- a/server/recceiver/cfstore.py
+++ b/server/recceiver/cfstore.py
@@ -1308,13 +1308,11 @@ def push_to_cf(
     _log.info("Pushing updates for %s begins...", ioc_info)
     count = 0
     sleep = 1.0
-    success = False
-    while not success and (processor.cf_config.push_always_retry or count < processor.cf_config.push_max_retries):
+    while processor.cf_config.push_always_retry or count < processor.cf_config.push_max_retries:
         count += 1
         try:
             update_method(processor, record_info_by_name, records_to_delete, ioc_info)
-            success = True
-            return success
+            return True
         except RequestException as e:
             _log.error("ChannelFinder update failed: %s", e)
             retry_seconds = min(60, sleep)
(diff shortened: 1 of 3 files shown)


server/recceiver/cfstore.py
```diff
@@ -1300,9 +1306,13 @@ def poll(
         ioc_info: The IOC information.
     """
     _log.info("Polling for %s begins...", ioc_info)
+    count = 0
     sleep = 1.0
     success = False
-    while not success:
+    while not success and (
+        count < processor.cf_config.polling_retry_count or processor.cf_config.polling_infinite_retries
```
simon-ess: Somewhat pedantic, but I would invert the order of this `or` conditional for readability.

anderslindho:
I exposed this as configurable in #146 @jacomago 

Nvm dropped it

server/recceiver/cfstore.py
simon-ess, line 1316: This could just be a `return True` and we could even drop the `success` variable (with a `return False` at the end...)

server/recceiver/cfstore.py
```diff
@@ -1299,19 +1305,19 @@ def poll(
...
+            return True
         except RequestException as e:
             _log.error("ChannelFinder update failed: %s", e)
             retry_seconds = min(60, sleep)
             _log.info("ChannelFinder update retry in %s seconds", retry_seconds)
             time.sleep(retry_seconds)
             sleep *= 1.5
-    _log.info("Polling %s complete", ioc_info)
-    return success
+    _log.error("Pushing updates for %s complete, failed after %d attempts", ioc_info, count)
```
simon-ess, line 1322: This message is a bit odd "complete" but "failed after" seem at odds with each other.

simon-ess (approved):

PR #146, opened 2026-05-04
Mixed bugfixes

anderslindho:
See individual commits for descriptions. In particular the NActive issue seems highly relevant for critical issues we have in production at ESS, for our largest sub-network (~2500 IOCs).


server/recceiver/recast.py
```diff
@@ -387,6 +387,8 @@ def isDone(self, P, active):
 
     def buildProtocol(self, addr):
         active = self.NActive < self.maxActive
+        if active:
+            self.NActive += 1
```
simon-ess: I would have this as a companion `else` to the `if not active` below 🔧

server/recceiver/cfstore.py
```diff
@@ -1289,29 +1291,37 @@ def poll(
     record_info_by_name: Dict[str, RecordInfo],
     records_to_delete,
     ioc_info: IocInfo,
+    max_retries: int = 0,
```
simon-ess: This is not used in this commit.

anderslindho: Dropped the commit since another PR covered the same config change

jacomago (approved):

PR #148, opened 2026-05-05
fix(processors): add missing getint to ConfigAdapter (#145 regression)

anderslindho:
#145 introduced a breaking bug, which was not caught by CI because the entire test suite consists of Docker integration tests - there were no tests exercising the configuration parsing path.

We will need an immediate new patch release (1.9.4) after this MR.

### DIFF
--- a/server/tests/test_config.py
+++ b/server/tests/test_config.py
@@ -1,0 +1,136 @@
+import textwrap
+from configparser import ConfigParser
+from pathlib import Path
+
+from recceiver.cfstore import CFConfig, CFProcessor
+from recceiver.processors import ConfigAdapter, ProcessorController
+
+
+def make_adapter(section: str = "cf", values: dict = None, env: dict = None) -> ConfigAdapter:
+    parser = ConfigParser()
+    parser.add_section(section)
+    for key, value in (values or {}).items():
+        parser.set(section, key, str(value))
+    adapter = ConfigAdapter(parser, section)
+    if env:
+        adapter.env_vars = env
+    return adapter
+
+
+class TestConfigAdapterGet:
+    def test_returns_value(self):
+        adapter = make_adapter(values={"baseurl": "https://cf:8080"})
+        assert adapter.get("baseurl") == "https://cf:8080"
+
+    def test_returns_default_when_missing(self):
+        adapter = make_adapter()
+        assert adapter.get("baseurl", "fallback") == "fallback"
+
+    def test_returns_none_when_missing_and_no_default(self):
+        adapter = make_adapter()
+        assert adapter.get("baseurl") is None
+
+    def test_env_var_overrides_config(self):
+        adapter = make_adapter(values={"baseurl": "https://original"}, env={"baseurl": "https://override"})
+        assert adapter.get("baseurl") == "https://override"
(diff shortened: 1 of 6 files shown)


server/tests/test_config.py
```diff
@@ -0,0 +1,136 @@
...
+    def test_returns_default_on_invalid_value(self):
+        adapter = make_adapter(values={"pushmaxretries": "notanint"})
+        assert adapter.getint("pushmaxretries", 10) == 10
+
+    def test_env_var_overrides_config(self):
+        adapter = make_adapter(values={"pushmaxretries": "5"}, env={"pushmaxretries": "99"})
+        assert adapter.getint("pushmaxretries", 10) == 99
+
+
+class TestCFConfigLoads:
```
jacomago: should probably have this in test_cfstore.py

server/tests/test_processors.py
```diff
@@ -0,0 +1,136 @@
+import textwrap
```
jacomago, line 1: Can we split out these tests from the docker tests.

i.e. make the docker tests in an intergration-tests folder and then can split up the github ci to allow fail on integration-tests but not on unit tests.

jacomago (approved):

server/tests/test_processors.py
anderslindho: Won't do it now, but I'll create a generic task in our backlog to improve on this entire test setup mess

server/tests/test_config.py
anderslindho: Changed the structure a bit

jacomago (approved):

server/tests/test_processors.py
```diff
@@ -0,0 +1,99 @@
...
+        config_file.write_text(
+            textwrap.dedent(
+                """\
+                [recceiver]
+                procs = cf
+
+                [cf]
+                pushMaxRetries = 5
+                """
+            )
```
jacomago, line 94: I get a fail on this one

 instname = plugname
    
>           plug = plugs[plugname]
E           KeyError: 'cf'

server/recceiver/processors.py:102: KeyError

anderslindho: Did you `pip install .` first?

jacomago: I think the problem is plugin.getPlugins(interfaces.IProcessorFactory) requires twisted being started up for this test.

jacomago: i.e. as in twisted/recceiver_plugin you have

cffactory = processors.ProcessorFactory("cf", cfstore.CFProcessor)

anderslindho: This worked for me locally and seems to work in CI. I am wondering if you just are missing some prereq.

anderslindho: > I think the problem is plugin.getPlugins(interfaces.IProcessorFactory) requires twisted being started up for this test.

I think you are wrong. Nothing in the docs for `getPlugins` suggests that the reactor must be running.

anderslindho: Assuming your issues are local and closing this. We can return to fix if this turns out to be false.

PR #152, opened 2026-05-06
fix(tests): refresh CF client after container restart

anderslindho:
Docker's ephemeral port mapping reassigns the host port whenever a container is stopped and started. The cf_client fixture captured the original port at class setup; after restart_container / start_container it pointed at a dead port, causing every wait_for_sync attempt to fail silently until the 180-second timeout expired.

Recreating the client from the compose state after each CF restart picks up the new port and lets the tests pass.

### DIFF
--- a/server/tests/test_single_ioc.py
+++ b/server/tests/test_single_ioc.py
@@ -13,6 +13,7 @@ from .client_checks import (
     channels_match,
     check_channel_property,
     create_client_and_wait,
+    create_client_from_compose,
     wait_for_sync,
 )
 from .docker import (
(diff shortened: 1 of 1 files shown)


jacomago (approved):
Impressive find!

server/tests/test_single_ioc.py
simon-ess, line 79: Do we need `cf_client` as a parameter now? (same applies to other usage)

anderslindho: This PR makes no change to functional reqs, it only explicitly reinits the client for these tests to pass.

PR #153, opened 2026-05-06
Fix recCeiver issues

anderslindho:
See individual commits

### DIFF
--- a/server/recceiver/cfstore.py
+++ b/server/recceiver/cfstore.py
@@ -1107,9 +1107,9 @@ def _update_channelfinder(
 
     if iocid not in iocs:
         _log.warning(
-            "IOC %s did not send an initial transaction to join IOC list: %s",
+            "IOC %s did not send an initial transaction to join IOC list (%d IOCs known)",
             ioc_info,
-            [(ioc.ioc_name, ioc.ioc_id) for ioc in iocs.values()],
+            len(iocs),
         )
 
     if ioc_info.hostname is None or ioc_info.ioc_name is None:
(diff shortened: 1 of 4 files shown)


server/recceiver/recast.py
```diff
@@ -255,7 +255,7 @@ def __repr__(self):
 
 class CollectionSession(object):
     timeout = 5.0
-    trlimit = 0
+    trlimit = 5000
```
jacomago, line 258: Should we make this configurable?

server/recceiver/cfstore.py
```diff
@@ -1107,9 +1107,9 @@ def _update_channelfinder(
 
     if iocid not in iocs:
         _log.warning(
-            "IOC %s did not send an initial transaction to join IOC list: %s",
+            "IOC %s did not send an initial transaction to join IOC list (%d IOCs known)",
             ioc_info,
```
jacomago, line 1126: Maybe should have the first 3 IOCs, just for help debugging with small number of IOCs.

jacomago (commented):
Could we tests for the non-logging changes?

jacomago:
Tried to compare against my per ioc lock branch https://github.com/jacomago/recsync/tree/ioc-per-lock .
Looks like could still do both.

anderslindho:
> Tried to compare against my per ioc lock branch https://github.com/jacomago/recsync/tree/ioc-per-lock. Looks like could still do both.

Yeah, I think the per-IOC lock really is a better solution than this, but this is the more minimal fix that hopefully would settle our production issues @jacomago

server/recceiver/cfstore.py
anderslindho: TBH I am not sure about that. I feel like that's us trying to improve logging in one place when we really are interested in logging in one or more other places.

server/recceiver/recast.py
anderslindho: It is already configurable since before - commitSizeLimit

jacomago (approved):

PR #154, opened 2026-05-06
refactor(tests): organize by test category for easier filtering and parallel CI runs

anderslindho:


server/twisted/plugins/dropin.cache
jacomago, line 1: Dont think you want this file.

jacomago (commented):
Probably want to update github actions to run the tests separately.

server/twisted/plugins/dropin.cache
anderslindho: Done x2

jacomago (approved):

PR #157, opened 2026-05-07
Fix sonar issues for server/recCeiver (only)

anderslindho:


server/tests/integration/docker_compose.py
```diff
@@ -84,7 +84,7 @@ def clone_container(
 ) -> str:
     container_id = container_id or compose.get_container(host_name).ID
     if not container_id:
-        raise Exception("Container not found")
+        raise RuntimeError("Container not found")
```
jacomago, line 87: Maybe could be more explicit.

server/recceiver/mock_client.py
```diff
@@ -1,6 +1,8 @@
 from requests import HTTPError
 from twisted.internet.address import IPv4Address
 
+MOCK_CF_HTTP_ERROR = "Mock Channelfinder Client HTTPError"
```
jacomago, line 6: Maybe better to make a specific exception that extends httperror

server/recceiver/mock_client.py
```diff
@@ -11,59 +11,58 @@ def __init__(self):
...
+        ]
+
+    def _find_by_names(self, names):
+        return [self.cf[name] for name in names if name in self.cf]
+
+    def _find_active(self):
+        return [
+            channel
+            for channel in self.cf.values()
+            if any(prop["name"] == "pvStatus" and prop["value"] == "Active" for prop in channel["properties"])
```
jacomago: use enums for some of these. Would be good to split up cfstore into 

- cfstore
  - store (actual processors)
  - recc_cf_client (all methods used to push to cf)
  - model (conversion between types, and types)

or something. then can import from model maybe.

.github/workflows/server.yml
```diff
@@ -41,8 +41,14 @@ jobs:
             python-version: ${{ matrix.python-version }}
         - name: Install
           run: |
-            python -m pip install --upgrade pip
-            pip install .
+            python - <<'PY'
```
jacomago: this feels ugly, couldn't it be simpler in pure bash?

server/tests/integration/docker_compose.py
anderslindho: It is more explicit than it was :)

server/recceiver/mock_client.py
anderslindho: You do realise that I am doing the bare minimum to get our sonar score better, right?

jacomago: I move mean you can import from cfstore to get the "Active". so this would be simpler.

server/recceiver/mock_client.py
anderslindho: I agree, but I think that's out of scope here

jacomago (approved):

PR #159, opened 2026-05-08
feat(server): observability improvements and startup/retry fixes           

anderslindho:
Addresses several operational pain points found in production at ESS.

**Fixes**                                                                                                                            
 - `push_always_retry` defaulted to `True` (a regression), which could freeze all commit processing indefinitely when CF was unreachable. Default is now `False`.                                                                                               
 - `cleanOnStart`/`cleanOnStop` swept channels synchronously on the reactor thread, blocking IOC commits for the full duration of the sweep. Both now run in a background thread.                                                                                      
 - Warning messages for IOC connection problems were ambiguous and could fire twice for the same event. A numeric iocName (ephemeral source port) is now flagged explicitly.
                                                                                                                                   
**Observability**                                                                                                                  
 - Periodic status lines (configurable via statusInterval, default 60 s) log active/queued connections against `maxActive` and the CF processor's IOC and channel counts - the key indicators that were invisible during the production incident.                   
 - CF push duration is logged per attempt, making CF latency regressions visible in the log stream.            
 - Optional Prometheus metrics endpoint (metricsPort, requires `pip install recceiver[metrics]`). Exposes connection state, CF processor state, and a push duration histogram. Degrades gracefully to no-ops if prometheus-client is not installed.

### DIFF
--- a/server/demo.conf
+++ b/server/demo.conf
@@ -126,5 +126,6 @@
 # Number of times to retry polling before giving up. Default is 10.
 #pushMaxRetries = 10
 
-# Whether to retry polling indefinitely until success. Default is True.
-#pushAlwaysRetry = True
+# Whether to retry polling indefinitely until success. Default is False.
+# Enabling this holds the global CF commit lock until CF recovers; use with caution.
+#pushAlwaysRetry = False
--- a/server/demo.conf
+++ b/server/demo.conf
@@ -37,6 +37,10 @@
 # Interval in seconds between periodic status log lines (0 to disable)
 #statusInterval = 60.0
 
+# TCP port to expose Prometheus metrics on (0 or absent to disable).
+# Requires prometheus_client: pip install recceiver[metrics]
+#metricsPort = 0
+
 # Idle Timeout for TCP connections.
 #tcptimeout = 15.0
 
--- a/server/demo.conf
+++ b/server/demo.conf
@@ -34,6 +34,9 @@
 # Time interval for sending recceiver advertisments
 #announceInterval = 15.0
 
+# Interval in seconds between periodic status log lines (0 to disable)
+#statusInterval = 60.0
+
 # Idle Timeout for TCP connections.
 #tcptimeout = 15.0
 
--- a/server/pyproject.toml
+++ b/server/pyproject.toml
@@ -36,6 +36,7 @@ dependencies = [
   "twisted>=22.10,<23; python_version<'3.8'",
   "twisted>=24.11,<24.12; python_version>='3.8'",
 ]
+optional-dependencies.metrics = [ "prometheus-client" ]
 optional-dependencies.test = [ "pytest>=8.3,<8.4", "pytest-cov>=6,<7", "testcontainers>=4.8.2,<4.9" ]
 urls.Repository = "https://github.com/ChannelFinder/recsync"
 
--- a/server/recceiver/application.py
+++ b/server/recceiver/application.py
@@ -11,6 +11,7 @@ from zope.interface import implementer
 
 from twisted import plugin
 
+from . import metrics
 from .announce import Announcer
 from .processors import ProcessorController
 from .recast import CastFactory
@@ -54,6 +55,7 @@ class RecService(service.MultiService):
 
         self.port = int(portn or "0")
         self.statusInterval = float(config.get("statusInterval", "60.0"))
+        self.metricsPort = int(config.get("metricsPort", "0"))
 
         for addr in config.get("addrlist", "").split(","):
             if not addr:
@@ -113,11 +115,22 @@ class RecService(service.MultiService):
         # This will start up plugin Processors
         service.MultiService.privilegedStartService(self)
 
+        if self.metricsPort > 0:
+            if metrics.available:
+                self.reactor.listenTCP(self.metricsPort, metrics.make_site(), interface=self.bind)
+                _log.info("Prometheus metrics available on port %d", self.metricsPort)
+            else:
+                _log.warning("metricsPort configured but prometheus_client is not installed; metrics disabled")
+
+        metrics.connections_limit.set(self.tcpFactory.maxActive)
+
         if self.statusInterval > 0:
             self._statusLoop = task.LoopingCall(self._logStatus)
             self._statusLoop.start(self.statusInterval, now=False)
 
     def _logStatus(self):
+        metrics.connections_active.set(self.tcpFactory.NActive)
+        metrics.connections_waiting.set(len(self.tcpFactory.Wait))
         _log.info(
             "status: connections active=%d/%d queued=%d",
             self.tcpFactory.NActive,
--- a/server/recceiver/application.py
+++ b/server/recceiver/application.py
@@ -4,7 +4,7 @@ import logging
 import random
 
 from twisted.application import service
-from twisted.internet import defer, pollreactor
+from twisted.internet import defer, pollreactor, task
 from twisted.internet.error import CannotListenError
 from twisted.python import log, usage
 from zope.interface import implementer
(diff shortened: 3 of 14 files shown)


server/recceiver/cfstore.py
```diff
@@ -382,10 +385,18 @@ def _stop_service_with_lock(self):
         """Stop the CFProcessor service with lock held.
 
         If clean_on_stop is enabled, mark all channels as inactive.
+        The sweep runs in a thread so the reactor stays live during shutdown.
```
jacomago: We don't want reactor live during any type of clean do we?

server/recceiver/cfstore.py
```diff
@@ -619,6 +627,12 @@ def _commit_with_thread(self, transaction: CommitTransaction):
 
         record_info_by_name = CFProcessor.record_info_by_name(record_infos, ioc_info)
         self.update_ioc_infos(transaction, ioc_info, records_to_delete, record_info_by_name)
+        if not transaction.connected and ioc_info.ioc_id not in self.iocs:
+            _log.warning(
```
jacomago: Shouldn't we exit here?

server/demo.conf
```diff
@@ -34,6 +34,9 @@
 # Time interval for sending recceiver advertisments
 #announceInterval = 15.0
 
+# Interval in seconds between periodic status log lines (0 to disable)
```
jacomago, line 37: not 100% sure about this, but ok.

And also update recceiver-full.conf.

server/recceiver/cfstore.py
```diff
@@ -1366,19 +1366,26 @@ def push_to_cf(
         records_to_delete: The records to delete.
         ioc_info: The IOC information.
     """
-    _log.info("Pushing updates for %s begins...", ioc_info)
+    _log.info("CF push start: %s (%d channels)", ioc_info, len(record_info_by_name))
     count = 0
     sleep = 1.0
     while processor.cf_config.push_always_retry or count < processor.cf_config.push_max_retries:
+        if not processor.running:
```
jacomago: Is there a way to have a test for this?

server/pyproject.toml
```diff
@@ -32,6 +32,7 @@ dependencies = [
   "requests",
   "twisted",
 ]
+optional-dependencies.metrics = [ "prometheus-client" ]
```
jacomago, line 39: Is there an open-telemetry alternative so we can support more than prometheus?

server/recceiver/metrics.py
```diff
@@ -0,0 +1,89 @@
...
+        ["result"],
+        registry=_registry,
+    )
+    cf_commit_duration_seconds = Histogram(
+        "recceiver_cf_commit_duration_seconds",
+        "CF push duration in seconds",
+        buckets=[0.1, 0.5, 1.0, 2.5, 5.0, 10.0, 30.0, 60.0],
+        registry=_registry,
+    )
+
```
jacomago, line 51: Maybe we could also expose some of the configuration parameters? That way can compare directly in grafana to them.

server/pyproject.toml
anderslindho: Probably, but 1) I have never worked with it, and 2) prometheus seemed a minimal reasonable opt-in that fits our site and current needs perfectly. OpenTelemetry makes sense for recCeiver replacements.

server/recceiver/metrics.py
anderslindho: Strong dislike. Doesn't fit our needs at all, and is arguably an anti-pattern - at the very least it mixes concerns. For our case we use ansible and are in control of configuration. It should not slide. This PR adds live monitoring of changing values, which config values are not.

jacomago: You're actually already exposing the one I thought of, maxActive.

anderslindho: FWIW current active / max active was the entire reason for the exporter

server/recceiver/cfstore.py
```diff
@@ -587,7 +587,15 @@ def _commit_with_thread(self, transaction: CommitTransaction):
         ioc_name = transaction.client_infos.get("IOCNAME")
         if not ioc_name:
             ioc_name = str(port)
-            _log.debug("IOC at %s:%d did not send IOCNAME; using port as iocName", host, port)
+            _log.debug("IOC at %s:%d has no iocName; using source port as iocName", host, port)
+        if ioc_name.isdigit() and 1024 <= int(ioc_name) <= 65535:
```
simon-ess: This condition seems odd to me. Does this ever occur outside of the conditional right above it?

server/recceiver/application.py
```diff
@@ -110,9 +111,24 @@ def privilegedStartService(self):
...
+            "status: connections active=%d/%d queued=%d",
+            self.tcpFactory.NActive,
+            self.tcpFactory.maxActive,
+            len(self.tcpFactory.Wait),
+        )
+
     def stopService(self):
         _log.info("Stopping RecService")
 
+        if hasattr(self, "_statusLoop") and self._statusLoop.running:
```
simon-ess: Ugh, I don't love using `hasattr`. Why not just initialise `_statusLoop` to None?

server/recceiver/cfstore.py
```diff
@@ -298,6 +298,12 @@ def startService(self):
         finally:
             self.lock.release()
 
+        self._statusLoop = task.LoopingCall(self._logStatus)
+        self._statusLoop.start(60.0, now=False)
```
simon-ess: This doesn't look very configurable to me...

server/recceiver/cfstore.py
anderslindho: The phrasing has been (/will be) changed which hopefully makes the intent clearer

server/recceiver/cfstore.py
anderslindho: Yes, it can. The intent is to catch any iocName that looks like an ephemeral source port, whichever way it got there. But I guess you mean to drop the range comp?

simon-ess: Yes and no. On the one hand ports will always be number in that range unless something super weird is going on, but I guess my main question was mor about if it is ever the case that:
* ioc_name is _not_ None
* ioc_name is a stringified integer

This would clearly be a client error, but does that happen?

anderslindho: The file has since been deleted and the code moved, so this diff is outdated. To your question: yes, it can fire even when IOC name came from the IOC/reccaster - if someone misconfigured the reccaster to send a numeric string as IOCNAME. It's pathological but possible, so the check is intentionally placed outside the if not block.

So it prob does not add a huge amount of value, and I can simplify if you want.

server/demo.conf
```diff
@@ -37,6 +37,10 @@
 # Interval in seconds between periodic status log lines (0 to disable)
 #statusInterval = 60.0
 
+# TCP port to expose Prometheus metrics on (0 or absent to disable).
+# Requires prometheus_client: pip install recceiver[metrics]
+#metricsPort = 0
```
jacomago, line 42: Missing this config in recceiver_full.conf.

server/recceiver/recast.py
```diff
@@ -278,6 +278,7 @@ def flush(self):
             return
 
         transaction, self.transaction = self.transaction, Transaction(self.ep, id(self))
+        self.transaction.client_infos = dict(transaction.client_infos)
```
jacomago, line 281: Does this really work with an IOC?

anderslindho: Fix is only applicable for large IOCs, with more than 5k (trlimit) records. The copy to subsequent transaction applies also to smaller IOCs but does not matter for those as they will flush all (incl client_infos) on first transaction. But for >5k, the subsequent transaction starts with empty client infos.

Keep in mind that reccaster proto is ordered: (all) AddInfo goes before AddRecord.

jacomago (approved):

PR #164, opened 2026-05-13
Fix/code quality issues

anderslindho:


server/tests/unit/cf/test_config.py
```diff
@@ -28,6 +28,16 @@ def test_default_push_always_retry(self):
         config = CFConfig.loads(adapter)
         assert config.push_always_retry is True
 
+    def test_default_env_owner_variable(self):
+        adapter = make_adapter()
+        config = CFConfig.loads(adapter)
+        assert config.env_owner_variable == "ENGINEER"
+
+    def test_env_owner_variable_from_config(self):
+        adapter = make_adapter(values={"envownervariable": "RESPONSIBLE_ENGINEER"})
```
simon-ess, line 39: Can I just state how annoying the mixed cases are?

server/recceiver/interfaces.py
```diff
@@ -22,23 +19,11 @@ class ITransaction(Interface):
     recid: {'key':'val'}
     """)
 
+    aliases = Attribute("A dict mapping record id to list of alias names")
 
-@dataclass
-class SourceAddress:
-    host: str
-    port: int
```
simon-ess, line 29: Aren't we losing something here? Previously we had a concrete data structure for the source address. Now we do not.

The same is true I suppose of all the properties: we are going from typed properties to non-typed ones, aren't we?

server/pyproject.toml
```diff
@@ -50,7 +50,7 @@ line-length = 120
 # Enable Pyflakes (`F`) and a subset of the pycodestyle (`E`)  codes by default.
 # Unlike Flake8, Ruff doesn't enable pycodestyle warnings (`W`) or
 # McCabe complexity (`C901`) by default.
-lint.select = [ "E", "F", "I" ]
+lint.select = [ "E", "F", "G", "I" ]
```
simon-ess, line 52: Update the comment above?

server/recceiver/cf/config.py
jacomago, line 1: Ah oops by me.

server/recceiver/application.py
jacomago, line 1: Yeah, this is a confusing one. I think this is necessary or not necessary depending on how you launch the application.

server/recceiver/interfaces.py
jacomago, line 1: I added these so I can get the type checking in vscode.... I can see why maybe they shouldn't be here...

jacomago (approved):

server/tests/unit/cf/test_config.py
anderslindho: I don't know, can you? ;)

server/recceiver/interfaces.py
anderslindho: We are, yes, but this was just a hack. No instance of the class was ever used.

anderslindho: I tried another way to do the same, maybe you can try it out @jacomago

PR #165, opened 2026-05-15
(feat): per-IOC locking

anderslindho:


server/tests/unit/test_recast.py
```diff
@@ -13,6 +14,47 @@ def _make_session() -> CollectionSession:
...
+
+        session.factory.commit.side_effect = mock_commit
+
+        session.ioc_info("IOCNAME", "IOC1")
+        session.flush()
+        session.close()
+
+        assert len(committed) == 2
+        assert committed[-1].connected is False
+        session.proto.transport.loseConnection.assert_called_once()
```
simon-ess, line 120: Can these be combined with a parameterisation? 🔧

server/recceiver/cf/processor.py
```diff
@@ -163,57 +172,67 @@ def _stop_service_with_lock(self):
...
+    def _commit_with_lock(self, transaction: interfaces.ITransaction, iocid: str) -> defer.Deferred:
+        self._cancelled[iocid] = False
+        try:
+            result = yield deferToThread(self._prepare_commit, transaction, iocid)
+            if result is None:
+                return  # disconnect-before-upload: nothing to push
+            ioc_info, record_info_by_name, records_to_delete, iocs_snap, ciids_snap = result
+            yield self._push_to_cf_async(iocid, record_info_by_name, records_to_delete, ioc_info, iocs_snap, ciids_snap)
+        except defer.CancelledError:
+            raise
```
simon-ess: Should we log this instead of just re-raising it?

server/recceiver/cf/processor.py
```diff
@@ -163,57 +172,67 @@ def _stop_service_with_lock(self):
...
         return d
 
+    @defer.inlineCallbacks
+    def _commit_with_lock(self, transaction: interfaces.ITransaction, iocid: str) -> defer.Deferred:
+        self._cancelled[iocid] = False
+        try:
+            result = yield deferToThread(self._prepare_commit, transaction, iocid)
+            if result is None:
+                return  # disconnect-before-upload: nothing to push
+            ioc_info, record_info_by_name, records_to_delete, iocs_snap, ciids_snap = result
```
jacomago, line 227: Maybe should make this a dataclass.

server/recceiver/cf/processor.py
```diff
@@ -300,10 +319,12 @@ def update_ioc_infos(
         if transaction.initial:
             self.iocs[iocid] = ioc_info
         if not transaction.connected:
-            records_to_delete.extend(self.channel_ioc_ids.keys())
+            records_to_delete.extend(self._ioc_channels.get(iocid, set()))
         for record_name in record_info_by_name:
-            self.channel_ioc_ids[record_name].append(iocid)
-            self.iocs[iocid].channelcount += 1
+            if iocid not in self.channel_ioc_ids[record_name]:
```
jacomago, line 326: This suggests maybe we should be using a set instead of a list for channel_ioc_ids[record_name]. And then I think there is a method for doing add if not exists.

server/tests/unit/cf/test_processor.py
```diff
@@ -115,30 +128,181 @@ def test_orphans_channel_absent_from_local_state(self):
...
+    def test_prune_removes_state_after_ioc_disconnects(self):
+        proc = make_processor()
+        iocid = f"{_HOST_A}:5064"
+        proc._ioc_locks[iocid] = DeferredLock()
+        proc._cancelled[iocid] = False
+        proc._ioc_channels[iocid].add("CHAN:1")
+        # iocid is NOT in proc.iocs → IOC has disconnected
+
+        proc._prune_ioc_state(None, iocid)
+
```
simon-ess: Semantically, isn't this comment more linked to the the following line isntead of the previous one? 🔧

server/recceiver/recast.py
```diff
@@ -361,34 +361,34 @@ class CastFactory(protocol.ServerFactory):
     maxActive = 3
 
     def __init__(self):
-        # Flow control by limiting the number of concurrent
-        # "active" connectons  Active means dumping lots of records.
-        # connections become "inactive" by calling isDone()
+        # Throttle concurrent uploading connections to control CF commit load.
+        # "Active" means currently uploading records; connections become
+        # "inactive" via isDone() once the upload completes.
         self.NActive = 0
```
simon-ess, line 367: Commenting here because I can't comment on the commit message. The commit message says that `maxActive now operate per-IOC` and that the conf files have been updated but:
* what does that even mean?
* This is not true; the only change to the conf files is about the `pushAlwaysRetry`.

server/recceiver/cf/processor.py
```diff
@@ -414,12 +414,14 @@ def _prepare_commit(self, transaction: interfaces.ITransaction, iocid: str):
...
-            # covers everything _handle_channel_is_old can look up without
-            # cloning the full channel map on every commit.
+            # Snapshot only the channels and IOCs this commit touches; that covers
+            # everything _handle_channel_is_old / create_default_properties can look
+            # up without cloning the full maps on every commit.
             channels_to_snapshot = set(records_to_delete) | self._ioc_channels.get(iocid, set())
             ciids_snap = {k: list(self.channel_ioc_ids[k]) for k in channels_to_snapshot if k in self.channel_ioc_ids}
+            needed_ioc_ids = {iid for ch in channels_to_snapshot for iid in self.channel_ioc_ids.get(ch, [])}
+            needed_ioc_ids.add(iocid)
+            iocs_snap = {iid: self.iocs[iid] for iid in needed_ioc_ids if iid in self.iocs}
```
simon-ess: This is fine, but I think that the commit message is misleading; it doesn't copy the dictionary, but creates a new one that references the old values. They are not copies, meaning that while it is true that the operation is O(number of IOCs) it isn't as serious as it sounds.

That said, I guess this is fine but it a little bit reeks of premature optimisation. We replace a likely heavily optimised call to `dict(other dict)` to "let's iterate by hand over some stuff and make a new one". Do we have evidence that this is an actual bottleneck speed-wise?

server/recceiver/application.py
```diff
@@ -75,6 +75,12 @@ def __init__(self, config):
     def privilegedStartService(self):
         log.info("Starting RecService")
 
+        # Each active IOC commit holds a thread for the CF HTTP call; size the
+        # pool so all maxActive commits can run concurrently (+ headroom for
+        # clean_service and other thread users). Twisted's default of 10 would
+        # otherwise throttle commits even when maxActive slots are available.
+        self.reactor.suggestThreadPoolSize(self.maxActive + 10)
```
simon-ess: I love the name "suggest" here 😁

server/tests/unit/test_recast.py
anderslindho: Maybe, but I don't think that would be right. They are not performing the same test with different inputs, but testing two different behaviours.

simon-ess: I dunno. They only diffrent as I see it is that if one of a list of exceptions is raised then basically the same thing happens; the only difference otherwise between the two is the last line.

That said, there might be some subtlety that I am missing.

server/recceiver/application.py
jacomago: 10 should be a constant or configurable.

jacomago (commented):
I'm wondering if we can get better testing for this. 

I'm thinking we could make the multi-ioc container tests could have maxActive = 2, since there are 4 iocs. 
Would also be nice to have an IOC with more channels than the commitMax which is defaulted to 5000, could simply set it to 1 and make sure tests work?

server/tests/unit/test_recast.py
anderslindho: I suspect you're seeing small diff -> parametrise, while I am seeing semantically different things -> do not. In either case, I don't think they should be joined, and you don't seem to feel strongly about this, so I will resolve.

server/recceiver/cf/processor.py
anderslindho: I have avoided dataclasses since we support py3.6 (to some degree, at least).

server/recceiver/cf/processor.py
anderslindho: Note quite that simple... `_handle_channel_is_old` does `channel_ioc_ids[channel_name][-1]` to get the last known IOC for a channel, so this is not a trivial change.

server/recceiver/cf/processor.py
anderslindho: > That said, I guess this is fine but it a little bit reeks of premature optimisation. We replace a likely heavily optimised call to dict(other dict) to "let's iterate by hand over some stuff and make a new one". Do we have evidence that this is an actual bottleneck speed-wise?

You're right that the operation most likely is less performant per iteration, but I think you may be missing the main reason for the change and the implications of it - lock contention. `_state_lock` is held by up to `maxActive`(=50 for us) concurrent threads, each running `_prepare_commit` simultaneously. At our scale - ~2500+ IOCs for the worst recCeiver - we are copying 2500+ entries per commit, while the replacement copies ~1-2 references per typical commit.

You're also right to call out that the foundation of this change is not completely data driven (as in by profiling), but it _is_ driven by observations at scale.

server/recceiver/recast.py
anderslindho: Updated

server/recceiver/cf/processor.py
jacomago: Already using dataclasses: https://github.com/ChannelFinder/recsync/blob/master/server/recceiver/cf/model.py

jacomago: But otherwise could use nameddict for example.

anderslindho: > Already using dataclasses

I know, just did not want to make it worse.

TBH I think I prefer to leave it as is - it seems at least OK now.

server/recceiver/cf/processor.py
simon-ess: I mean, it's doing a copy of 2500 references which I doubt is actually that slow; again, these aren't deep copies. I suppose your point is that there is a single state lock that means that each commit blocks each other commit, which is fair.

However, your last statement is entirely contradictory: without data, how is this fix addressing any observations at scale? The fact that things slow down doesn't mean that this is the bottleneck.

IMO without that data this change here should be done separately, and tested appropriately.

anderslindho: Dropped it for now

jacomago (commented):
Still think we should have more integration tests where the ioc count is purposefully above the configuration maxActive etc. Maybe even just alter the configuraiton for the multi-ioc tests to have maxActive =1 .

anderslindho:
> Still think we should have more integration tests where the ioc count is purposefully above the configuration maxActive etc. Maybe even just alter the configuraiton for the multi-ioc tests to have maxActive =1 .

Sorry, forgot to reply to this before @jacomago. (Always annoying how to deal with the summary comments in GH which aren't threads...)

I don't think we should be further extending the current "e2e" tests run in this project's CI. In fact, I think we should be making an effort to remove them (which actually is intended to partially be enabled by this MR).

I can, however, accept trying to increase coverage in a more isolated "mock-y" way, if that's what you feel is needed.

jacomago:
> > Still think we should have more integration tests where the ioc count is purposefully above the configuration maxActive etc. Maybe even just alter the configuraiton for the multi-ioc tests to have maxActive =1 .
> 
> Sorry, forgot to reply to this before @jacomago. (Always annoying how to deal with the summary comments in GH which aren't threads...)
> 
> I don't think we should be further extending the current "e2e" tests run in this project's CI. In fact, I think we should be making an effort to remove them (which actually is intended to partially be enabled by this MR).
> 
> I can, however, accept trying to increase coverage in a more isolated "mock-y" way, if that's what you feel is needed.

Ok, not the e2e tests, I agree they should be removed. Then I think tests that explicitly have maxActive and commitAmount (or whatever the variable is called) being low so that its clear the behaviour is still correct and can handle multiple IOCs.

jacomago (approved):

PR #168, opened 2026-06-05
Fix/1.9.6 prod observations

anderslindho:
This MR fixes a bunch of issues seen in production at ESS when deploying recCeiver 1.9.6 (and CF 5.1.0) with cleanOnStart active and freshly wiped DB.

- Expected ~9 M channels; CF showed ~4.75 M total, ~1.75 M active
- Active channel count climbed to ~2.1 M then dropped sharply to ~1.55 M within minutes of startup; never recovered
- recceiver-feb (largest instance, ~2500 IOC network): `known_iocs` ca 100 during incident; after manual restart climbed to expected level then crashed
- recceiver-ps: dropped from 387 to ~30 known IOCs in ~90 s mid-incident; cause unconfirmed (possible container restart, but not verified)
- 2 recceivers presenting 0 IOCs - there was, however, potentially 0 IOCs on those networks
- saw in logs many channelCount 0 - mainly maybe for feb

From prometheus exporter:
```
# HELP recceiver_connections_active Active uploading IOC connections
# TYPE recceiver_connections_active gauge
recceiver_connections_active -1.0
```

Also in logs:
```
2026-06-03T12:55:27+0000 [-] INFO:recceiver.application status: connections active=-60328/20 queued=0
```

Note that this MR adds a recceiver-clean utility. This is because we at ESS have decided to not use cleanOn* anymore - this is not, and never was, recCeiver's scope. It was a band-aid which we do not want to attempt fixing any further. We will instead use the utility as needed, and try to integrate better mechanisms in a future CF version or any potential CF replacement.

### DIFF
--- a/server/recceiver/cf/processor.py
+++ b/server/recceiver/cf/processor.py
@@ -648,7 +648,16 @@ class CFProcessor(service.Service):
     ) -> None:
         """Channel exists in CF but not in this commit — re-assign to its last known IOC."""
         last_ioc_id = self.channel_ioc_ids[cf_channel.name][-1]
-        cf_channel.owner = self.iocs[last_ioc_id].owner
+        last_ioc_info = self.iocs.get(last_ioc_id)
+        if last_ioc_info is None:
+            log.warning(
+                "Last IOC %s for channel %s not in local state; orphaning channel",
+                last_ioc_id,
+                cf_channel.name,
+            )
+            self._orphan_channel(cf_channel, ioc_info, channels, record_info_by_name)
+            return
+        cf_channel.owner = last_ioc_info.owner
         cf_channel.properties = _merge_property_lists(
             create_default_properties(ioc_info, recceiverid, self.channel_ioc_ids, self.iocs, cf_channel),
             cf_channel,
@@ -663,7 +672,10 @@ class CFProcessor(service.Service):
                     alias_channel = CFChannel(alias_name, "", [])
                     if alias_name in self.channel_ioc_ids:
                         last_alias_ioc_id = self.channel_ioc_ids[alias_name][-1]
-                        alias_channel.owner = self.iocs[last_alias_ioc_id].owner
+                        last_alias_ioc_info = self.iocs.get(last_alias_ioc_id)
+                        if last_alias_ioc_info is None:
+                            continue
+                        alias_channel.owner = last_alias_ioc_info.owner
                         alias_channel.properties = _merge_property_lists(
                             create_default_properties(
                                 ioc_info, recceiverid, self.channel_ioc_ids, self.iocs, cf_channel
--- a/server/recceiver/cf/processor.py
+++ b/server/recceiver/cf/processor.py
@@ -399,7 +399,21 @@ class CFProcessor(service.Service):
         self.update_ioc_infos(transaction, ioc_info, records_to_delete, record_info_by_name)
         poll_success = self._push_to_cf(record_info_by_name, records_to_delete, ioc_info)
         if not poll_success:
+            if transaction.connected:
+                self._evict_ioc(ioc_info.id)
             raise defer.CancelledError(f"Failed to commit transaction after polling retries: {transaction}")
+
+    def _evict_ioc(self, iocid: str) -> None:
+        """Remove an IOC from in-memory state so its next commit is treated as initial.
+
+        Called after a CF push exhausts retries: CF was never written, so
+        in-memory state would diverge from CF until the IOC reconnects.
+        """
+        # list() snapshots the set; remove_channel mutates it during iteration
+        for ch in list(self._ioc_channels.get(iocid, set())):  # NOSONAR
+            self.remove_channel(ch, iocid)
+        self.iocs.pop(iocid, None)
+        self._ioc_channels.pop(iocid, None)
 
     def remove_channel(self, record_name: str, iocid: str) -> None:
         """Unlink a channel from an IOC in channel_ioc_ids and decrement channelcount.
--- a/server/recceiver/cf/processor.py
+++ b/server/recceiver/cf/processor.py
@@ -1,6 +1,5 @@
 import datetime
 import logging
-import threading
 import time
 from collections import defaultdict
 from typing import Callable, Dict, List, Optional, Set
@@ -30,10 +29,6 @@ from recceiver.processors import ConfigAdapter
 log = logging.getLogger(__name__)
 
 
-class CFUpdateAbortedError(Exception):
-    """Raised when a CF push is abandoned after exhausting all retries."""
-
-
 @implementer(interfaces.IProcessor)
 class CFProcessor(service.Service):
     """IProcessor plugin that synchronises IOC record data to Channelfinder.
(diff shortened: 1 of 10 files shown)


server/recceiver/clean_tool.py
```diff
@@ -0,0 +1,96 @@
...
+
+def run_clean(cf_config: CFConfig, client=None, dry_run: bool = False) -> int:
+    """Mark all Active channels for cf_config.recceiver_id Inactive.
+
+    Returns the total count of channels swept. Pass a pre-built client for testing.
+    """
+    if client is None:
+        from channelfinder import ChannelFinderClient
+
+        client = PyCFClientAdapter(
```
jacomago, line 33: I think you can just move this into main? Then you don't need the client is None guard.

server/recceiver/clean_tool.py
```diff
@@ -0,0 +1,96 @@
...
+                BaseURL=cf_config.base_url,
+                username=cf_config.cf_username,
+                password=cf_config.cf_password,
+                verify_ssl=cf_config.verify_ssl,
+            ),
+            size_limit=int(cf_config.cf_query_limit),
+        )
+
+    total = 0
+    while True:
```
jacomago, line 48: Maybe we could pull this out and share the code with the other places we call clean?

server/recceiver/cf/processor.py
```diff
@@ -166,74 +158,62 @@ def stopService(self):
...
-            return d
+            return deferToThread(self.clean_service)
 
     def _start_background_clean(self):
         log.info("CF Clean: background startup sweep beginning")
         deferToThread(self.clean_service).addErrback(lambda err: log.error("CF Clean background sweep failed: %s", err))
 
-    def _get_ioc_lock(self, iocid: str) -> DeferredLock:
-        """Return the per-IOC DeferredLock, creating it on first use.
+    # @defer.inlineCallbacks # Twisted v16 does not support cancellation!
```
jacomago: We are actually on Twisted > 20 I think?

server/recceiver/cf/processor.py
jacomago, line 1: We have lost all the doc strings to the methods. Maybe we could have a commit to add new ones.

server/tests/unit/cf/test_processor.py
jacomago, line 1: Also lost a lot of tests here that I think are still relevant.

server/recceiver/clean_tool.py
anderslindho: Nah, we use it with tests. So then we'd need heavier monkey patching - this seems easier IMO.

server/recceiver/clean_tool.py
anderslindho: A shared abstraction would need to parameterise the retry behaviour... Leaving them separate.

tynanford:
> Note that this MR adds a recceiver-clean utility. This is because we at ESS have decided to not use cleanOn* anymore - this is not, and never was, recCeiver's scope. It was a band-aid which we do not want to attempt fixing any further. We will instead use the utility as needed, and try to integrate better mechanisms in a future CF version or any potential CF replacement.

That is interesting. What happened with these issues when you disabled cleanOnStop and cleanOnStart?

server/recceiver/clean_tool.py
jacomago: I thought it would make less monkey patching...

server/recceiver/clean_tool.py
```diff
@@ -0,0 +1,96 @@
...
+            " (dry-run)" if dry_run else "",
+        )
+        if not dry_run:
+            client.update_property(
+                CFProperty(CFPropertyName.PV_STATUS.value, cf_config.username, PVStatus.INACTIVE.value),
+                [ch.name for ch in channels],
+            )
+        total += len(channels)
+        if dry_run:
+            break
```
simon-ess, line 65: The dry run option + loop here seems very weird. I guess the point is that more channels could come active? Or is this a pagination thing?

server/recceiver/clean_tool.py
```diff
@@ -0,0 +1,96 @@
...
+
+        client = PyCFClientAdapter(
+            ChannelFinderClient(
+                BaseURL=cf_config.base_url,
+                username=cf_config.cf_username,
+                password=cf_config.cf_password,
+                verify_ssl=cf_config.verify_ssl,
+            ),
+            size_limit=int(cf_config.cf_query_limit),
+        )
```
simon-ess, line 41: Do we really need to have a `None` argument to test this? I get that you are allowing the tests to pass a mocked object, but can't you just mock the creation of the object in the test without passing it as an argument?

server/recceiver/cf/processor.py
```diff
@@ -648,7 +648,16 @@ def _handle_channel_is_old(
...
-        cf_channel.owner = self.iocs[last_ioc_id].owner
+        last_ioc_info = self.iocs.get(last_ioc_id)
+        if last_ioc_info is None:
+            log.warning(
+                "Last IOC %s for channel %s not in local state; orphaning channel",
+                last_ioc_id,
+                cf_channel.name,
+            )
+            self._orphan_channel(cf_channel, ioc_info, channels, record_info_by_name)
+            return
```
simon-ess: I wonder if this can be moved up to the calling function. This is a bit of a complicated jump-around of code, and since this block is a quick-exit that essentially jumps to the branch not chosen from the calling function, it seems like we could clean this up and make it a bit clearer.

anderslindho:
@tynanford we did not want to change the configuration, but instead immediately rolled back. I can thus only guess, but my guess would have been that the issues would have been mitigated if we had changed `cleanOnStart` - that the active/inactive reports would have been much more accurate.

server/recceiver/clean_tool.py
anderslindho: if client creation moves to `main()`, tests need to patch `ChannelFinderClient` and `PyCFClientAdapter`

server/recceiver/clean_tool.py
anderslindho: Pagination. I can add a comment if you want?

server/recceiver/clean_tool.py
anderslindho: We don't strictly need to, but it makes for less monkey patching. I'll resolve this thread as it's the same thing as discussed in https://github.com/ChannelFinder/recsync/pull/168#discussion_r3371626725.

server/recceiver/cf/processor.py
anderslindho: I think it can, yeah. It would be a new refactor commit, but I can try to do it if you think it significantly improves readability.

simon-ess: I think it would, yes. This is pretty spaghetti-ish code already and I think having a clearer trace through when and why what is happening would help.

server/recceiver/clean_tool.py
simon-ess: I think I would prefer that. It sort of makes the dry run a bit confusing otherwise, since it seems like the dry run won't actually give the correct results but only a first-pass-through count.

server/recceiver/cf/processor.py
simon-ess: I had been thinking something more like
```py
    try:
        last_ioc_id = self.channel_ioc_ids[cf_channel.name][-1]
        last_ioc_owner = self.iocs[last_ioc_id].owner
    except KeyError:
        self._orphan_channel()
    else:
        self._handle_channel_is_old()
```
which I feel is a lot easier to read and avoids such nested conditionals. But this is an improvement IMO

jacomago (approved):

PR #175, opened 2026-08-05
chore(server): bump version to 1.10.0

anderslindho:


tynanford:
maybe after this is merged if we decide to keep 3.6 around? https://github.com/ChannelFinder/recsync/pull/178

anderslindho:
@tynanford Good question. To be clear, you're suggesting we fix py3.6 support before releasing 1.10.0? If so, I think I agree.

Your branch will need to be rebased on origin/master to pick up the test-integration fix, so I can't approve quite yet. But if you can fix + approve this here PR then I can do both merges and create the release tomorrow - I have the release notes already prepared.

tynanford:
Yes exactly. I just rebased the py3.6 PR with your integration test fix. Looks like github CI actions are down - https://www.githubstatus.com/ . I'll try and re-run them today

tynanford (approved):

jacomago (approved):

Issue #177, opened 2026-08-05
requires-python declares 3.6, but the package needs 3.9


anderslindho:
The package claims to support Python 3.6 and above, but it does not run on
anything below 3.9. Installing it on 3.6, 3.7 or 3.8 works; importing it fails
immediately. The syntax responsible is in processors.py and recast.py, and is
3.9 or later.

The 3.6 and 3.7 jobs miss this because build-server only runs
"pip install --no-deps .", which never imports what it installed. The jobs that
do import recceiver all pin 3.9.

So: move the declared floor to 3.9, or rewrite the constructs to restore 3.6.
Either way the build job should import the package, so the claim is tested.

tynanford:
https://github.com/ChannelFinder/recsync/pull/178

PR #178, opened 2026-08-05
Restore python 3.6 and update CI to import module after installing

tynanford:
https://github.com/ChannelFinder/recsync/issues/177

This updates processors.py and recast.py to work with python3.6 and above - updated type hints and remove removeprefix which is >=3.9

Had claude help with updating the CI. It now actually imports the package as @anderslindho mentioned and manually installs the pip packages for 3.6 since setuptools on 3.6 doesn't work with the pyproject.toml setup. Claude also found an issue with the pyproject.toml file: The <3.8 and <3.7 lines for requests and twisted overlap for python 3.6.

### DIFF
--- a/.github/workflows/server.yml
+++ b/.github/workflows/server.yml
@@ -44,10 +44,27 @@ jobs:
             minor=$(python3 --version | cut -d. -f2)
             if [ "$minor" -ge 7 ]; then
               python -m pip install --upgrade "pip==24.0" --only-binary=:all:
+              python -m pip install .
             else
+              # Python 3.6 runs from a checkout (see README), not a pip install.
+              # Just install the runtime deps, pinned per pyproject.toml.
               python -m pip install --upgrade "pip==21.3.1" --only-binary=:all:
+              python -m pip install \
+                "channelfinder @ https://github.com/ChannelFinder/pyCFClient/archive/refs/tags/v3.2.0.zip#sha256=ec89537f336c4cb8db18909c4b02176dd5fbd03f5872a6527129879cbf3f7c85" \
+                "dataclasses==0.8" \
+                "requests==2.27.1" \
+                "twisted==21.7.0"
             fi
-            python -m pip install --no-deps .
+        - name: Import smoke test
+          run: |
+            minor=$(python3 --version | cut -d. -f2)
+            if [ "$minor" -lt 7 ]; then
+              # Intentional for 3.6: run from the checkout, as documented.
+              PYTHONPATH=. python -c "from twisted.plugins import recceiver_plugin"
+            else
+              # Run outside the checkout so this fails if pip didn't install it.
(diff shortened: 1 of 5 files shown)


anderslindho (changes requested):
Unfortunately this doesn’t work as-is - note `Successfully installed UNKNOWN-0.0.0` in the 3.6 build log.

We’ll have to reintroduce `setup.cfg`, move the `[project]` / package metadata out of `pyproject.toml`, pin setuptools to something that still runs on 3.6, and fix the smoke test: right now `PYTHONPATH=.` imports from the checkout even when install failed, so it can’t catch this. After a real `pip install .`, run the import from `/tmp` (or any dir outside `server/`) so it has to use the installed package. We should probably also set ruff to `py37` (it doesn’t support 36).

tynanford:
Ah, good catch. Instead of reverting back to setup.cfg I propose we just have the CI test the import without installing it as a package. Confirmed in a test commit here that the CI will fail for py3.6 - py3.8 with `set` keyword: https://github.com/ChannelFinder/recsync/actions/runs/31213348556

Sonarqube is complaining i think because we don't have pinned exact versions in the pyproject.toml file. To me it seems like something that could be dismissed?

anderslindho:
OK with the smaller fix. Still some issues, however:
- The new comment cites the README for 3.6 running from a checkout, but that instruction is not there; server/README.md only documents `PYTHONPATH=$PWD twistd` for local server testing, and README.md still says "It requires Python 3.6 or above".
- `requires-python = ">=3.6"` and the 3.6 classifier in `pyproject.toml` are now wrong.
- We should prob accept and comment on the Sonar issue on your branch before we merge.

tynanford:
Sounds good to me, updated the CI comment and README to make this clear.

.github/workflows/server.yml
```diff
@@ -44,10 +44,27 @@ jobs:
...
+              python -m pip install .
             else
+              # setuptools on 3.6 is too old for install with pyproject.toml
+              # So just install the runtime dependencies manually
               python -m pip install --upgrade "pip==21.3.1" --only-binary=:all:
+              python -m pip install \
+                "channelfinder @ https://github.com/ChannelFinder/pyCFClient/archive/refs/tags/v3.2.0.zip#sha256=ec89537f336c4cb8db18909c4b02176dd5fbd03f5872a6527129879cbf3f7c85" \
+                "dataclasses==0.8" \
+                "requests==2.27.1" \
+                "twisted==21.7.0"
```
anderslindho, line 56: Wondering if we should document this in the README, at least dataclasses?

anderslindho (approved):

.github/workflows/server.yml
tynanford: yes good point https://github.com/ChannelFinder/recsync/pull/180

PR #179, opened 2026-08-06
fix(tests): correct the IOC dbd path

anderslindho:
The integration IOC failed to start against the current `ghcr.io/channelfinder/reccaster:master` image: `dbLoadDatabase` could not open `/recsync/dbd/demo.dbd`, so no record types were registered and `iocInit` aborted with a bad DBD. Every integration test then waited out its sync timeout against an IOC that never cast.

Reccaster renamed its image build paths from `/recsync` to `/reccaster` after the repository split, and the `:master` tag was rebuilt on 2026-08-03. Since that tag moves, CI broke without any change here.


ralphlange (approved):
LGTM

